PR #5, opened 2015-08-21
defaultStorage.getAllKeys: typo in catch block

pascience:
Stumbled upon this while casually browsing the source - I guess this is an untested code path.


pascience:
Closed because I confused PR title and commit title. See https://github.com/rt2zz/redux-persist/pull/6

PR #6, opened 2015-08-21
persistStore: typo in catch block

pascience:
Stumbled upon this while casually browsing the source - I guess this is an untested code path.


rt2zz:
thanks :)

I have not figured out how to test redux-persist yet. I think maybe we need a testing middleware to invoke callbacks when actiosn come through or something?

PR #7, opened 2015-08-29
fix: warnIfSetError referencing undefined `key`

joeybaker:
Previously, `key` was `undefined`.


rt2zz:
thanks!

joeybaker:
Any chance you can publish a new version?

rt2zz:
just published .4.4 with the fix, although it just occured to me this will need to be done for warnIfRemoveError as well

PR #18, opened 2015-10-11
Remove localStorage var that clobbers the global

doxxx:
Declaring a var localStorage at the top level clobbers the global localStorage defined by the browser with `undefined`.

It wasn't clear what that line was trying to accomplish and it works without it.


rt2zz:
I will have to take a look, the goal is to define localStorage so that if it is undefined (e.g. in node) it does not through an error.

doxxx:
I think that would be better done by having a no-op storage implementation that is selected instead of the asyncLocalStorage implementation.

rt2zz:
Agreed. Thanks for the PR!

PR #39, opened 2015-11-17
Fix for Immutable.js warning

thisiskylierose:
When checking if data is an object using lodash methods, part of the process is to determine is an object is actually an array. lodash checks for the length property and this is not supported in Immutable.js - this fix prevents an warning being logged in modern browsers and a breaking error in IE9 or below.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "test:watch": "npm test -- --watch",
     "clean": "rimraf lib dist",
     "build": "babel src --out-dir lib",
-    "build:watch": "watch 'npm run build' ./src",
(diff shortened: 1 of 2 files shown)


rt2zz:
looks good :+1:

rt2zz:
I will cut a new release after I have a chance to manuall test (there are currently no immutablejs tests currently)

thisiskylierose:
Ok then. Thank you. Let me know if there are any breaking issues. I have to integrate this with production code so it's tested but would be good to get feedback... just in case.

rt2zz:
released 1.4.3

thisiskylierose:
Thanks very much.

PR #40, opened 2015-11-17
Fix AsyncStorage import in Readme

af:


rt2zz:
:+1:

af:
Thanks, I'm just getting started with redux-persist, but so far it's super slick! :+1:

PR #42, opened 2015-12-01
Fixed issue with checking for window object

thisiskylierose:
Was causing issues with build. This check will avoid instances of error with window not defined in node env


rt2zz:
x.x :+1:

PR #43, opened 2015-12-03
Change the way of checking for localStorage access

andreaspsson:
The check was not safe for ChromeExtensions. In the case of the user blocking
3rd party cookies the access of window.localStorage throws an exception, that
cased the load of the module to fail instead of fallbacking to the
"non window.localStorage" version. That leads to that the application loading this module crashes at startup.

This is how it looks when you try to access localStorage
screen shot 2015-12-03 at 15 38 28

Problem can be reproduced and verified by building a simple ChromeExtension that accesses window.localStorage and the enable "Block third party cookies and site data" under Content Settings in the Chrome settings menu

### DIFF
--- a/src/defaults/asyncLocalStorage.js
+++ b/src/defaults/asyncLocalStorage.js
@@ -7,7 +7,16 @@ const noLS = process && process.env && process.env.NODE_ENV === 'production'
     return null
   }
 
-var localStorage = typeof window === 'object' && typeof window.localStorage !== 'undefined'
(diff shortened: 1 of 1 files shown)


andreaspsson:
Missed to pass all the tests I see. Will update the pull request tonight

rt2zz:
no worry, just a few style changes I fixed

andreaspsson:
cool, thanks!

PR #45, opened 2015-12-14
Fix for rehydrating a top-level string field.

af:
Hi,

This is one change I needed to make after upgrading from `1.4.1`. I was rehydrating a root-level string field, and `reducedState[key]` was `null`. This caused `reducedState[key].size` to throw an error. Adding this extra `(reducedState[key] &&` guard fixed the issue for me.

### DIFF
--- a/src/autoRehydrate.js
+++ b/src/autoRehydrate.js
@@ -22,7 +22,9 @@ module.exports = function autoRehydrate (config = {}) {
   }
 
   function checkIsObject (data, reducedState, key) {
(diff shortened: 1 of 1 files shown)


af:
Left some extraneous braces in my first commit, just force-pushed a fixed version

rt2zz:
This looks good, but I want to add a test for this before cutting a new release. Feel free to write one, or I will later this week.

:+1:

af:
Cool, I'll look into that. Thanks!

rt2zz:
just release v1.5.0 with your patch & a test :+1:

af:
Awesome, thanks!

PR #54, opened 2016-01-28
"setImmediate() is not available" error

goodhyun:
In an environment such as nw.js, [global.setImmediate() is not linked as setImmediate()](https://github.com/nwjs/nw.js/issues/897).


rt2zz:
looks good. I need to figure out why travis is complaining about the unused var. I should be able to merge this tomorrow.

Issue #61, opened 2016-02-18
storage is undefined when using asyncSessionStorage


jonespen:
... because of missing `return` in asyncLocalStorage.js. Submitting a pull request

zalmoxisus:
It seems to be already fixed in `2.0`. Please try `npm i  redux-persist@next`.

PR #62, opened 2016-02-18
return sessionStorage

jonespen:
closes GH-61


rt2zz:
I will cut a new release later today

rt2zz:
Out in v1.5.6 and also 2.0.0-rc3

If you are interested in trying v2 `npm install redux-persist@next`

The main difference is there is one big REHYDRATE action instead of one per reducer.

PR #69, opened 2016-03-21
Add support for Mozilla's localForage library as a storage backend

Klathmon:
This is a really quick PR to allow support for Mozilla's localForage library to be used as a drop-in storage backend.

The only thing that is needed, is that the `getAllKeys(cb)` function in localforage is called just `keys(cb)`, so i added a quick-and-dirty check to see if the function 'keys' exists then "rename" it. I was needed in both the persistStore.js and getStoredState.js files, but it's a grand total of like 3 lines in each.

I also updated the readme to add a line about its support with localForage.

I have no issues breaking the check out into another file and maybe adding a check for all of the required functions and an error if any don't exist, but i didn't want to change too much without asking first (it's a pain when a PR comes in with like 100 lines changed and no explanation why...)

Let me know!

By the way, nice choice in standard :+1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ const secondaryPersistor = persistStore(store, {storage: specialBackupStorage, s
 ```
 
 ## Storage Backends
-**localStorage** (default), react-native **AsyncStorage**, Mozilla's **[localForage](https://github.com/mozilla/localForage) library** or a conforming **custom** storage api. Custom storage API should be an object with the following methods: `setItem` `getItem` `removeItem` `getAllKeys` each with the function signature as found in [react-native AsyncStorage](http://facebook.github.io/react-native/docs/asyncstorage.html#content).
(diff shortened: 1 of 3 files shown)


rt2zz:
Awesome. So this will use localForage if it exists, and if not default to localStorage? I am unclear how this is achieved, does localForage monkey patch localStorage?

Klathmon:
You pass in a reference to the 'localforage' variable/global as the 'storage' value in redux-persists config.

It's basically exactly the same as using AsyncStorage from react.

I was actually thinking about adding an example to the readme, but it was identical to the AsyncStorage one but with the word 'AsyncStorage' replaced with 'localforage'

Klathmon:
Hey I hate to be a pain but would you be able to cut another release (or prerelease) for this?

rt2zz:
@Klathmon sure, I am going to make some other changes and cut a v3 prerelease later today. In the meantime you can always point your npm to github i.e. `rt2zz/redux-persist`

Klathmon:
Thanks, I appreciate it.

And I'd point it at the repo but the lib dir isn't published, so I'd need to compile to es5 myself which I don't want to do.

rt2zz:
ah that is annoying :/ ill keep you updated

rt2zz:
just cut v3.0.0-prealpha `npm i redux-persist@next`

This includes a modified version of this PR, removes the action buffer which was too automagical and  sometimes caused problems, and now uses json-stringify-safe to prevent errors on cyclic objects.

PR #70, opened 2016-03-30
Update all dependencies and switch to babel 6

marvinhagemeister:
This PR updates all dependencies to the latest version. Most changes are necessary because babel changed quite a bit with version 6. `babel-preset-stage-2` is needed for the spread operator in `persistStore.spec.js`.

There are two stylistic changes needed to satisfy standardjs:

```js
// Wrong
const foo = bar || () => {};

// Right
const foo = bar || function () {};

// Wrong (missing parenthesis)
return next => foo;

// Right
return (next) => foo;
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "persist and rehydrate redux stores",
   "main": "lib/index.js",
   "scripts": {
-    "test": "standard && NODE_ENV=production mocha --compilers js:babel/register --recursive",
+    "test": "standard && NODE_ENV=production mocha --compilers js:babel-register --recursive",
     "test:watch": "npm test -- --watch",
     "clean": "rimraf lib dist",
     "build": "babel src --out-dir lib",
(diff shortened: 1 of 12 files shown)


marvinhagemeister:
I'll squash the commits into one, once I fix the CircleCI errors.

marvinhagemeister:
Seems like this is a problem with older node versions (pre v4). Looking at the log of CircleCI npm always installs babel v5 which is the root cause.

rt2zz:
@marvinhagemeister definitely not a requirement :) I will turn off circle and we can keep rolling with travis.

package.json
```diff
@@ -27,20 +27,27 @@
...
-    "babel-core": "^5.8.33",
-    "babel-eslint": "^4.1.3",
-    "expect": "~1.12.2",
+    "babel-cli": "^6.6.5",
+    "babel-core": "^6.7.4",
+    "babel-eslint": "^6.0.0",
+    "babel-plugin-transform-runtime": "^6.6.0",
+    "babel-preset-es2015": "^6.6.0",
+    "babel-preset-es2015-loose": "^7.0.0",
+    "babel-preset-stage-0": "^6.5.0",
```
rt2zz: Is stage-0 needed?

package.json
```diff
@@ -27,20 +27,27 @@
...
-    "babel": "^5.8.29",
-    "babel-core": "^5.8.33",
-    "babel-eslint": "^4.1.3",
-    "expect": "~1.12.2",
+    "babel-cli": "^6.6.5",
+    "babel-core": "^6.7.4",
+    "babel-eslint": "^6.0.0",
+    "babel-plugin-transform-runtime": "^6.6.0",
+    "babel-preset-es2015": "^6.6.0",
+    "babel-preset-es2015-loose": "^7.0.0",
```
rt2zz: What does loose do? If we need to change some code to be more "by the book" I am open to that.

package.json
```diff
@@ -27,20 +27,27 @@
...
+    "babel-register": "^6.7.2",
+    "babel-runtime": "^6.6.1",
+    "expect": "~1.16.0",
     "immutable": "^3.7.6",
-    "lodash": "^3.10.1",
-    "mocha": "~2.3.3",
-    "redux": "~3.0.4",
-    "rimraf": "~2.4.3",
-    "standard": "^5.3.1"
+    "lodash": "^4.6.1",
```
rt2zz: :+1:

src/index.js
```diff
@@ -1,8 +1,8 @@
-const createAsyncLocalStorage = require('./defaults/asyncLocalStorage')
+const createAsyncLocalStorage = require('./defaults/asyncLocalStorage').default
```
rt2zz: we should switch these to `import`s

rt2zz:
overall looks great! I made a few comments, but the main thing is if you can squash it down I will get this merged asap.

package.json
marvinhagemeister: I can check if loose mode is still needed. It was included in the `babel5` config as well. In a nutshell: output generated with loose mode can potentially be faster, but may lead to compatibility problems when transpiling won't be necessary in the future. More info: http://www.2ality.com/2015/12/babel6-loose-mode.html

package.json
marvinhagemeister: Will check tomorrow

rt2zz:
awesome stuff and much needed!

marvinhagemeister:
Glad I could help!

PR #74, opened 2016-04-14
Cover null cases in checkIfPlain()

af:
Some of my rehydrated state is null, and it's causing an error during rehydration. This change fixes things for me


af:
Thanks! Any chance you could push a point release for this? I can't install from git since the ES5 source is generated in a `prepublish` script

rt2zz:
Ah yes that is unfortunate. I will not be in a place to publish a release until Monday, but I will asap

af:
No problem, thanks!

PR #76, opened 2016-04-17
Fixed typo `combineReducers` -> `compose` in docs

joerx:
`combineReducers` isn't used anywhere, but `compose` is not declared. Assuming this was a typo?


rt2zz:
:+1:

PR #81, opened 2016-04-29
Avoid importing the whole lodash

ratson:


rt2zz:
Do you know why redux does not use babel-lodash? I based my implementation on that.

ratson:
Because the author avoided `import _ from 'lodash`, by importing used functions directly, e.g. `import isPlainObject from 'lodash/isPlainObject'`

rt2zz:
:+1:

jorilallo:
Thanks for making the change, it has a pretty significant effect on the bundle size. Would be nice to get a release as `lib` isn't part of the repo.

PR #86, opened 2016-05-05
Update README.md

sscaff1:
If you pass in null as the initialState of createStore, then you cannot define your reducer like this:

`export function myStore(state = {}, action) {` because initialState wins out. 
If initialState is undefined though, then the initial state in your reducer wins. 
This is simply a redux issue that I'm bringing up and hence making this PR. I'll reference an issue from Redux:
https://github.com/reactjs/redux/issues/1189


rt2zz:
👍

PR #96, opened 2016-05-13
Added initial state

dylancompanjen:
Added initial state which is necessary to do a valid conditional (if) statement. (Else you will get errors)


rt2zz:
I highly prefer the class property syntax, although I realize not everyone has that set up in there transpiler

👍

Issue #100, opened 2016-05-17
persistor object documentation


Moharu:
Today I had to develop a functionality that required a manual persistor rehydrate() call. After a while, I got it to work just fine, but I think it would be appreciated if the docs were improved to explain the rehydrate method and parameters (such as that it need a serialized input and can receive a callback). I don't know if this is kind of implicit, as I'm just developing my first project with redux, but I think it would help a lot.

Anyways, props for the awesome project :smile:

Moharu:
I also ended up using the "getStoredState" method that is available in the API. It was really easy to understand and use, but a mention in the docs would be useful

rt2zz:
👍 thanks!

PR #101, opened 2016-05-19
Rehydrate method docs (in persistor object)

Moharu:
just a draft
(https://github.com/rt2zz/redux-persist/issues/100)


rt2zz:
thanks for this! it looks like you are working off of an older version of redux-persist though.

In v3 we support non-serialized state by default, and if you want to pass in serial state that requires `{serial: true}`

so e.g.:
```js
persistor.rehydrate({foo: 'bar'})
persistor.rehydrate(JSON.stringify({foo: 'bar'}), {serial: true})
```

https://github.com/rt2zz/redux-persist/blob/master/src/createPersistor.js#L63-L80

Moharu:
Oops, you are right, I was using v2.
Updating the PR :smile:

rt2zz:
:heart:

PR #103, opened 2016-05-24
Fix typo in autoRehydrate.js

igrayson:
While reading the file, it seemed this is what was intended?


rt2zz:
it is expected reducedState and state have the same shape. In this case though it might actually make more sense to spread reducedState instead of state...

Let me think this over

igrayson:
At this line, `reducedState === state`, so this proposed change cannot change behavior.

It simply reads more easily, to see us predicate on `state`, which the subsequent spread consumes, than `reducedState`.

Issue #64, opened 2016-02-23
Any way to use redux-persist with immutable Map as the entire state?


peterlazar1993:
Want to integrate redux-persist with this https://github.com/multunus/react-native-boilerplate. Maybe a recipe for immutable state

rt2zz:
I am interested in supporting top level immutable state. You can `JSON.stringify` a immutable Map to a plain JSON string correct? In this case all we have to worry about is transforming the raw object back to a Map on rehydrate and also how to take the entire state and map it out into per-reducer chunks.

I believe this will require a new option `mapStateToPartials` or something like that. There will probably need to be corresponding changes throughout the code base.

After typing this out, I believe it is possible to implement with no impact on performance, but it will take some work so no ETA at this time, although PR's are appreciated. If you are short on time I would probably recommend looking at https://github.com/michaelcontento/redux-storage which deals which avoids dealing with state partials and hence has a much simpler story around immutable support.

zalmoxisus:
If we omit autorehydrate, it should be easy implemented with [Alternate Usage](https://github.com/rt2zz/redux-persist#initialstate-rehydration) 

```js
getStoredState(persistConfig, (err, initialState) => {
  const store = createStore(reducer, Immutable.fromJS(initialState))
  persistStore(store, persistConfig)
  // ...
})
```

rt2zz:
that works very well for rehydration, but there is still an outstanding problem where persistStore expects the state object returned from getState to be a plain object. A trivial albeit non-ideal performance solution would be to call `Immutable.toJS(store.getState())` inside of our store subscriber.

zalmoxisus:
I guess if we do `toJS` every time, then we lose the performance gained by Immutable.js, and better not use it at all, unless we persist only some reducers.

peterlazar1993:
I see, not entirely related to this issue but does using an immutable structure as the top level state provide any value?

zalmoxisus:
@peterlazar1993, I think no signifiant value. Better to use the standard Redux `combineReducers` and use `Immutable.js` on the reducer basis.

Then we'll have something like 
```js
getStoredState(persistConfig, (err, initialState) => {
  const initialImmutableState = {};
  Object.keys(initialState).forEach(function(key) {
   initialImmutableState[key] = fromJS(obj[key]);
  });
  const store = createStore(reducer, initialImmutableState)
  persistStore(store, persistConfig)
  // ...
})
```

[...]

PR #113, opened 2016-06-08
Allow for any type of root level state

rufman:
This addresses issue (#64), when you have an immutable object as the root state. 

The `redux-persist-immutable-state` package provides the functions for actually iterating over an immutable root state as well as getting and setting keys.

Usage of redux-persist-immutable-state:

```
import { stateIterator, stateGetter, stateSetter, 
         stateReconciler, lastStateInit } from 'redux-persist-immutable-state';

persistStore(state, {
  transforms: [reduxPersistImmutable], 
  stateIterator: stateIterator,  
  stateGetter: stateGetter, stateSetter: stateSetter,
  lastStateInit: lastStateInit
})
```

I've spent some time benchmarking the persisting of new data on state changes, as well as rehydrating the state from localStorage. Here are the results:
chart

The skeleton around those benchmarks is here: https://github.com/rufman/benchmark-redux-persist
If you want to run the tests yourself just clone https://github.com/rufman/redux-persist into the `lib`  directory of `benchmark-redux-persist` and run `webpack`. Then `open index.html` and run the tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+# What about redux-persist?
+This is a fork of redux-persist published as `redux-persist-2` until the PR https://github.com/rt2zz/redux-persist/pull/113 gets merged.
+
 # Redux Persist
 Persist and rehydrate a redux store.
 
--- a/src/autoRehydrate.js
+++ b/src/autoRehydrate.js
@@ -2,6 +2,8 @@ import { REHYDRATE } from './constants'
 import isStatePlainEnough from './utils/isStatePlainEnough'
 
 export default function autoRehydrate (config = {}) {
+  config.stateReconciler = config.stateReconciler || defaultStateReconciler
+
   return (next) => (reducer, initialState, enhancer) => {
     return next(createRehydrationReducer(reducer), initialState, enhancer)
   }
@@ -19,26 +21,8 @@ export default function autoRehydrate (config = {}) {
 
         let inboundState = action.payload
         let reducedState = reducer(state, action)
-        let newState = {...reducedState}
 
-        Object.keys(inboundState).forEach((key) => {
-          // if initialState does not have key, skip auto rehydration
-          if (!state.hasOwnProperty(key)) return
-
-          // if reducer modifies substate, skip auto rehydration
-          if (state[key] !== reducedState[key]) {
-            if (config.log) console.log('redux-persist/autoRehydrate: sub state for key `%s` modified, skipping autoRehydrate.', key)
-            newState[key] = reducedState[key]
-            return
-          }
-
-          // otherwise take the inboundState
-          if (isStatePlainEnough(inboundState[key]) && isStatePlainEnough(state[key])) newState[key] = {...state[key], ...inboundState[key]} // shallow merge
-          else newState[key] = inboundState[key] // hard set
-
-          if (config.log) console.log('redux-persist/autoRehydrate: key `%s`, rehydrated to ', key, newState[key])
-        })
-        return newState
+        return config.stateReconciler(state, inboundState, reducedState, config.log)
       }
     }
   }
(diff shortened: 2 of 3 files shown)


rt2zz:
awesome 👍 

I like this general approach, but want to wait for some resolution to https://github.com/reactjs/redux/issues/1792 before committing to a specific api.

rufman:
OK, cool. In the mean time I have published this PR as [redux-persist-2](https://www.npmjs.com/package/redux-persist-2) for anyone that already wants to use it.

dpereira411:
where is reduxPersistImmutable coming from? from redux-persist-transform-immutable? can I have an example?

rufman:
`reduxPersistImmutable` is from [redux-persist-immutable](https://www.npmjs.com/package/redux-persist-immutable), which serializes ImmutableJS substates. If your root level state is an ImmutableJS object you'll need to use `redux-persist-immutable-state`, but if it's a plain object and just your sub states are Immutable objects `redux-persist-immutable` will be all you need.

dpereira411:
thanks, so whats the difference between redux-persist-imutable and redux-persist-transform-immutable? github redirects redux-persist-immutable to redux-persist-transform-immutable

rt2zz:
@dpereira411 redux-persist-immutable is deprecated, as redux-persist-transform-immutable works with redux-persist v3+

mattkrick:
Could i get a little help on this API?
Still not getting it to persist from localStorage:

```
import {persistStore, autoRehydrate} from 'redux-persist-2';
import immutableTransform from 'redux-persist-transform-immutable';
import { stateIterator, stateGetter, stateSetter,
  stateReconciler, lastStateInit } from 'redux-persist-immutable-state';

persistStore(store, {
    transform: [immutableTransform],
    stateIterator: stateIterator,
    stateGetter: stateGetter,
    stateSetter: stateSetter,
    lastStateInit: lastStateInit
  });
```

rufman:
`transform: [immutableTransform]` should be `transforms: [immutableTransform]`

mattkrick:
that'll do it, thanks!

dpereira411:
shouldn't that be  `transforms: [immutableTransform()]` ?

kristojorg:
@rufman would you mind explaining or providing a quick example of how to use with top level immutable state. 

The only reason I have top-level immutable state is to be able to use the performant immutable deepMerge. I don't implement a combinereducers and instead have a custom root reducer since some actions (such as api responses) will have affects across many parts of the state tree. 

Could you also describe exactly what you're doing? 

Thanks!

ntucker:
I am getting this to persist, but it doesn't seem to be handling the rehydrate action properly. The action contains the data, but it's not altering the state.


I looked at https://github.com/rufman/redux-persist-immutable-state/blob/master/src/index.js and noticed the statereconciler is not made to work with immutablejs:

```
Object.keys(inboundState).forEach(
```

that line is for JS Objects, not immutable maps.

ntucker:
autoRehydrate doesn't seem to be working

```
import { persistStore, autoRehydrate } from 'redux-persist-2'
import immutableTransform from 'redux-persist-transform-immutable'
import { stateIterator, stateGetter, stateSetter,
         stateReconciler, lastStateInit } from 'redux-persist-immutable-state'

  let enhancer = compose(
    autoRehydrate(),
    applyMiddleware(
      ...middleware,
    ),
  )

  const store = createStore(rootReducer, Map(), enhancer)
  debug('rh:persist')(settings.persistOptions)
  persistStore(store, {
    whitelist: ['counter'],
    debounce: 500,
    storage: localforage,
    transforms: [immutableTransform({})],
    stateIterator,
    stateGetter, stateSetter,
    lastStateInit, stateReconciler,
  })
```

It saves state fine (I checked the store), but it won't initialize. It creates an action properly with all the data, but that doesn't update the state for some reason. I put some console.log in the autorehydrate enhancer and I noticed its reducer is only called once on the @@INIT action, then never again. It's not called on the persist/REHYDRATE or anything else. Anyone know why that might be?

marcellosachs:
@ntucker I ran into a similar issue - basically you have to pass `stateReconciler` as a config option to `autoRehydrate` instead of to `persistStore` : 

```
import { persistStore, autoRehydrate } from 'redux-persist-2'
import immutableTransform from 'redux-persist-transform-immutable'
import { stateIterator, stateGetter, stateSetter,
         stateReconciler, lastStateInit } from 'redux-persist-immutable-state'

  let enhancer = compose(
    autoRehydrate({stateReconciler}),
    applyMiddleware(
      ...middleware,
    ),
  )

  const store = createStore(rootReducer, Map(), enhancer)
  persistStore(store, {
    whitelist: ['counter'],
    debounce: 500,
    storage: localforage,
    transforms: [immutableTransform({})],
    stateIterator,
    stateGetter, stateSetter,
    lastStateInit, 
  })

ntucker:
Thanks @marcellosachs !

rufman:
Yeah, I guess I need to go over the docs and add that. My bad.

ntucker:
FYI @rt2zz : #158 still has the redux-persist-2 stuff in the README

> +# What about redux-persist?
> +This is a fork of redux-persist published as `redux-persist-2` until the PR https://github.com/rt2zz/redux-persist/pull/113 gets merged.

rt2zz:
@ntucker yikes good catch 🙌

Jbarget:
if anyone could cast an eye at the issue below i'd be very grateful! I'm having some issues with my set up

https://github.com/rt2zz/redux-persist-transform-immutable/issues/15

ntucker:
The reconciler problem wasn't my issue. The problem is the reducer from createRehydrationReducer is only run once from the @@INIT action.

stateReconciler only matters INSIDE that reducer, but it doesn't even get the chance to run.

PR #119, opened 2016-06-14
editing purge and purgeAll to return thenable

eclifford:
I ran into the situation where I needed to know when the underlying calls to `removeItem` were finished executing (since they are async). My PR simply propagates the thenable resolution once all `removeItem`'s resolve. With this update calls to `purge` and `purgeAll` are thenable.

### DIFF
--- a/src/createPersistor.js
+++ b/src/createPersistor.js
@@ -86,20 +86,22 @@ export default function createPersistor (store, config) {
 
   function purge (keys) {
     if (typeof keys === 'undefined') {
-      purgeAll()
+      return purgeAll()
     } else {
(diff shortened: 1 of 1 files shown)


rt2zz:
This is definitely desired functionality.

Concerns:
1) should also support callbacks
2) I dont think the purgeAll implementation works as it relies on getAllKeys returning a promise

rt2zz:
@eclifford thanks for the work! I adapted it and applied it to the coming v4 PR. 

Regarding my comments, I have decided callbacks is a feature that can be added later, and I updated the default getAllKeys implementation to support promises.

PR #121, opened 2016-06-17
return if error getAllKeys applied with error

konradk:


rt2zz:
👍 
This makes sense, but should probably return regardless of NODE_ENV. Also tests are failing because of standard style.

PR #140, opened 2016-07-26
Adds optional config to announce persist operations as actions

caiobalthazar:
Adds optional config to announce persist operations as actions. 
Batches the persisted keys in single announcement. Debouces batches instead of single keys.


rt2zz:
my bad for not reviewing this sooner. Is there a reason for closing? it is no longer applicable?

If you still think this is valuable can you explain why the `PERSISTED` action is useful? Could you instead rely on `REHYDRATE` as a proxy for `PERSISTED`?

caiobalthazar:
Hey there, the reason I closed was because I realized the persistor blacklist was meant for store keys and I had used it thinking that it was meant for action types. I would have to remove those 3 lines and submit a new request.

In my case, the optional PERSISTED action would be useful for two main things:
  1- To update any user interface widget telling the user that there are no pending changes to be persisted;
  2- To allow the application to sync its (debounced) persisted state with debug tools.

PR #146, opened 2016-08-09
Guarded getAllKeys, should return on error and do not proceed further…

### LINKED ISSUE #122: LocalStorage not accessible
Users can have localStorage disabled which makes set/get throw and there is no guard against that scenario.

Andarist:
… to throw at next line. Also improved localStorage detection.

I have embedded my app in a webview on Android, and in a normal use case it has `window.localStorage` but it is a null. So such situation should not be IMHO treated as having a localStorage available. Personally I would just shorten the check from

`typeof window.localStorage !== 'undefined' && window.localStorage !== null`

to just `window.localStorage` but wanted to follow your convention here.

Fixes #122

### DIFF
--- a/src/defaults/asyncLocalStorage.js
+++ b/src/defaults/asyncLocalStorage.js
@@ -10,7 +10,7 @@ const noStorage = process && process.env && process.env.NODE_ENV === 'production
 
 function hasLocalStorage () {
   try {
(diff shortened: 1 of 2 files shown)


rt2zz:
@Andarist I agree a simple `window.localStorage` check is probably best

Also good call on handling getAllKeys, I will get this merged asap

PR #149, opened 2016-08-16
Update README.md

pellejacobs:
Small change to the readme:

-  `import constants from 'redux-persist/constants'` will just return `undefined` as there seems to be no default export in `constants.js`

- update `Transforms` part to make clear reducers are passed in


rt2zz:
Thanks!

one note: it second argument to the transform functions is not action but rather reducer key. I will fix in a follow up commit

Issue #127, opened 2016-06-24
Static purgeAll


nevir:
It'd be handy to have a static `purge` and `purgeAll` that can be called before constructing a persistor.

I'm looking to be able to do this after fatal errors that have likely landed the user in a bad state (and it's safer to just wipe the store, have them log in again, and rebuild their state)

rt2zz:
this is available in the 4.0.0-alpha1 release 🎈

nevir:
Sweet, thanks!

rt2zz:
note: it is exported as `purgeStoredState`

`purgeStoredState(config)` is the equivalent to "purgeAll"

PR #158, opened 2016-08-28
v4

rt2zz:
- support immutable state ( 🎈  thanks @rufman )
- return Promises in default storage engines
- return Promise from purge methods

@rufman thanks for the great work on supporting immutable from #113 and I apologize for taking so long to get it merged in! since it requires a full 5 extra config options, I propose the following: Lets release a new module `redux-persist-immutable` which wraps redux-persist and proxies in the necessary config as well as well as the `redux-persist-transform-immutable` transform.

Towards this end, I renamed the config options from e.g. `config.stateSetter` to `config._stateSetter`, since the intended usage is via a proxy module (or a higher order module if you will).

Let me know your thoughts. I am still working on a few more tweaks and adding more tests in preparation for v4, I hope to cut a release this week.

cc/ @corbt @marvinhagemeister @rufman

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,6 @@
+# What about redux-persist?
+This is a fork of redux-persist published as `redux-persist-2` until the PR https://github.com/rt2zz/redux-persist/pull/113 gets merged.
+
 # Redux Persist
 Persist and rehydrate a redux store.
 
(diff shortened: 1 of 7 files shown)


rt2zz:
Also this adds promise support for purge #119 and static purgeStoredState export #127

rufman:
@rt2zz Yeah I think not exposing any of those options directly is a good idea. I'll look at starting an implementation of a "higher order library" this week.

PR #171, opened 2016-09-18
Add umd bundle via webpack. Fixes #170

### LINKED ISSUE #170: Create UMD dist
Is it possible to get a UMD dist? Similar to what Redux and Redux-Thunk provides.

marvinhagemeister:
The umd bundle is working, but the bundle size seems to be way bigger than expected. Will check again tomorrow.

### Features

- umd build + minified umd build
- es build (for webpack 2 or rollup)
- support for unpkg

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,11 @@
 {
-  "presets": ["es2015", "stage-2"]
+  "presets": ["stage-2", ["es2015", { "modules": false }]],
+  "env": {
+    "commonjs": {
+      "presets": ["es2015"]
+    },
+    "es": {
+      "plugins": ["./tools/use-lodash-es"]
+    }
+  }
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 lib/
+dist/
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 lib/
 dist/
+es/
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "babel-preset-stage-2": "^6.5.0",
     "cross-env": "^2.0.1",
     "immutable": "^3.7.6",
+    "lodash-webpack-plugin": "^0.10.1",
     "redux": "^3.5.2",
     "rimraf": "~2.5.2",
     "standard": "^8.0.0",
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "test": "standard && ava",
     "test:watch": "npm test -- --watch",
     "clean": "rimraf lib dist",
-    "build": "babel src --out-dir lib",
+    "build": "babel src --out-dir lib && webpack && NODE_ENV=production webpack",
     "build:watch": "npm run build ./src -- -watch",
     "prepublish": "npm run clean && npm run build"
   },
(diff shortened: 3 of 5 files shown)


rt2zz:
@marvinhagemeister probably because I removed the babel-plugin-lodash in favor of redux style `import something from 'lodash/something'`

Simplest solution might be to add webpack-lodash-plugin into the mix, but I defer to you on best implementation there.

rt2zz:
also ! awesome 🎈

rt2zz:
also in order to play nicely with https://unpkg.com/#/ I think we need to add
`"browser": "./dist/redux-persist.min.js"` to package.json
and add ./dist to package.json `"files"` array

marvinhagemeister:
@rt2zz I think you had the right intuition! It indeed caused by lodash.

I'll try to build the umd with rollup tomorrow. The tree-shaking should automatically take care of that. At work we have quite a few projects where that alone cut down the bundle size noticeably. We could even provide an es build via `js:next` for easier tree-shaking (only works with es6 syntax.

That is a great idea with unpkg! Should be easily doable 👍

marvinhagemeister:
got the es6 build working (for rollup or webpack2) but it's still a mystery why there is so much lodash stuff. We'll investigate further

marvinhagemeister:
Found the issue! The root cause is `lodash/forEach`which adds **12kb** to our minified bundle.

| build option | size minified (no gzip) |
|---|---|
| with `lodash/forEach` | 41kb |
| without `lodash/forEach` | 29kb |

There are 2 uses of `lodash/forEach` in our codebase:

### `createPersistor.js`

I'm unsure about a replacement with this one since we can't be sure what we are receiving. I'm not familiar with `ImmutableJS` or similar libraries, otherwise we could probably swap this with something else.

### `getStoredState.js`

This one seems to be totally save to replace with the native `Array.prototype.forEach` method.

@rt2zz What are your thoughts about this?

.babelrc
```diff
@@ -1,3 +1,11 @@
 {
-  "presets": ["es2015", "stage-2"]
+  "presets": ["stage-2", ["es2015", { "modules": false }]],
```
rt2zz: does modules: false not break based on the fact that we use es modules in the src/? Also I am curious why is modules: false necessary to get the umd build working?

rt2zz:
awesome! I am curious, you mentioned rollup & webpack 2, but I only see webpack 1, did you mean in the future we can consider switching?

.babelrc
marvinhagemeister: This is necessary for the es-build. Tree shaking only works with es2015 import statements because they are statically structured unlike commonjs where you can require modules inside an if-statement for example.

For our commonjs build I did use the normal `es2015` preset scoped to the `BABEL_ENV=commonjs` environment variable. That way we can easily switch between different build outputs.

Originally i had put the `modules: false` setting under the `es` env-var, but somehow babel throws an exception there. This smells like a bug in the config parser of babel itself.

marvinhagemeister:
You're right that the umd-bundles are created with webpack 1. What I mean with webpack 2 and rollup support is that they need to operate on `import` statements for tree-shaking to have any effect.

rt2zz:
nice 🎈  LGTM, is this ready to ship?

marvinhagemeister:
@rt2zz yep, this is ready to ship 🎉

rt2zz:
sick money https://unpkg.com/redux-persist@4.0.0-alpha5/dist/redux-persist.min.js

marvinhagemeister:
@rt2zz sweet! 🎉

PR #175, opened 2016-09-23
Hotfix: localForage -> getAllKeys assinging

willmcclellan:
Assigning the additional key to the storage object kills the some of the localForage functionality, because the prototype methods are not carried over. I don't see any benefit from assigning `getAllKeys` anyway, so this shouldn't affect anything else.

For a bit of background on the problems we were having... On Safari the storage adapter was falling back to webSQL which then failed because it lost some of it's functionality, particularly self.ready was undefined.

`TypeError: self.ready is not a function. (In 'self.ready()', 'self.ready' is undefined)`

https://github.com/localForage/localForage/blob/master/src/drivers/websql.js#L151

### DIFF
--- a/src/createPersistor.js
+++ b/src/createPersistor.js
@@ -22,7 +22,9 @@ export default function createPersistor (store, config) {
 
   // storage with keys -> getAllKeys for localForage support
   let storage = config.storage || createAsyncLocalStorage('local')
-  if (storage.keys && !storage.getAllKeys) storage = {...storage, getAllKeys: storage.keys}
+  if (storage.keys && !storage.getAllKeys) {
+    storage.getAllKeys = storage.keys
+  }
 
   // initialize stateful values
   let lastState = stateInit


marvinhagemeister:
@rt2zz LGTM, but you are way more familiar with this part.

rt2zz:
agreed, both ways are a little messy but this is a better approach.

willmcclellan:
Thanks for merging :) Any idea when this will be published to npm?

willmcclellan:
Also, do you guys have a workflow for installing from github? Currently can't deploy the fix off this branch because it requires npm prerelease workflow.

rt2zz:
just released 4.0.0-alpha6

no great workflow for working off master, due to the build requirement. Open to ideas for how to improve this workflow.

marvinhagemeister:
@willmcclellan Installing an npm package from a github repository [is supported by npm](https://docs.npmjs.com/files/package.json#git-urls-as-dependencies).

rt2zz:
@marvinhagemeister the challenge I have found is you still need to manually "npm run build" the project after it is installed from github since /lib is not checked in.

marvinhagemeister:
@rt2zz oh I see, didn't think of that! Perhaps we could run our build step as a `post-install` script. What do you think?

Issue #173, opened 2016-09-21
disable persistor?


TanninOne:
Couldn't find this in the documentation: Is there a way to stop a persistor created with createPersistor?

In the code I see the persistor subscribes a lambda and ignores the return value from store.subscribe which could be used to unsubscribe.

I see that I can apparently pause the persistor but I don't like the idea of having an unknown number of nops subscribed to the store.
Wouldn't it be nicer if the unsubscribe function was saved, _pause_ calls the unsubscribe and _resume_ subscribes again?

rt2zz:
Do you want to disable the persistor and re-enable at a later time? Or do you want to essentially destroy the persistor? 

In the former case, I think the existing `pause` functionality is sufficient. There will only ever be 1 subscription per persistor, and the performance cost of unsubscribing and resubscribing is negligible / not necessarily more performant depending on the circumstance.

TanninOne:
It's the latter, I want to completely remove/destroy the persistor. I'm not that worried about performance, it just feels 'dirty' to leave subscribers with no functionality hanging around.

rt2zz:
for sure, 
a) I agree lets add a "destroy" method. Not sure about naming, since it is technically more than just unsubscribing.
b) Curious what your use case is? This sounds interesting, are you doing something unorthodox like hot swapping apps?

TanninOne:
Not exactly hot-swapping apps but part of it. My app has (will have) a "mode" switch. In each mode (could be a few dozen but each user will probably only ever use a few) the app works with a different set of data and about half my application state refers to the current mode.
I don't want to holding the state of all modes in available all the time for various reasons (i.e. it would be harder to access the right one), instead I have one subtree for the current mode state and most of the code would be oblivious to the fact the mode can be changed. For that part of the state I want to use a second persistor (using whitelisting).
When the user changes the mode I would "detach" the persistor and activate a new one, replacing the whole mode-specific subtree, while the remaining state (window layout, language and so on) remains the same.
This operation doesn't have to be instantaneous, the user knows stuff is going to happen when he clicks the button.

I hope this is not a stupid approach, I'm fairly new to redux.

rt2zz:
sounds interesting and highly doable with redux. I will not be in a position to implement a destroy method anytime soon, but will gladly merge a pr that adds it to the persistor.

TanninOne:
Will do a PR, give me a few days.

PR #183, opened 2016-09-30
add support for unsubscribing a persistor

TanninOne:
This adds a function to the persistor that allows unsubscribing. It supports a callback which will be called after the last scheduled write  to the storage is complete.
See issue #173


TanninOne:
sorry, will fix code style

PR #184, opened 2016-09-30
add support for unsubscribing a persistor

TanninOne:
This adds a function to the persistor that allows unsubscribing. It supports a callback which will be called after the last scheduled write to the storage is complete.
See issue #173

second try

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 node_modules
-lib/
 dist/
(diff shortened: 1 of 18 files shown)


rt2zz:
closing due to staleness. The redux-persist code base is much more stable now, so I suspect adding a destroy method to the persistor could work.

PR #186, opened 2016-10-04
Prefer import over require in recipes.md

jasonmerino:
Just a quick documentation change proposal to prefer import over require


rt2zz:
👍

PR #197, opened 2016-10-19
WIP: use storage.clear when available

zhuangya:
`persistor.purge()` is kind of slow, i suggest if the storage implemented a `clear` method, just use it.

for now, `localStorage`, `sessionStorage`, `AsyncStorage` and `localForage` all have `clear()`

### DIFF
--- a/src/purgeStoredState.js
+++ b/src/purgeStoredState.js
@@ -10,14 +10,20 @@ export default function purgeStoredState (config, keys) {
 
   if (typeof keys === 'undefined') { // if keys is not defined, purge all keys
     return new Promise((resolve, reject) => {
-      storage.getAllKeys((err, allKeys) => {
-        if (err && process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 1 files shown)


rt2zz:
clear will remove all keys, where as purge removes only redux-persist keys. I think there is no way to safely do this. 

Is the slowness of purge problematic for your use case?

zhuangya:
> where as purge removes only redux-persist keys

hmm, didn't manage to think about that.

currently purge need 

- get all keys(with or without pattern)
- remove all keys in step one.

that's something like `O(n^2)`

so, that's kind of slow :(

rt2zz:
ya :( I agree its slow. Open to suggestions.

zhuangya:
maybe create something like plugin to do this?

users can `clear()` the storage with their own awareness in this way.

rt2zz:
@zhuangya as most users do not depend on fast purge, I think use cases that require this would be best served implementing purge themselves. i.e. calling `localStorage.clear()` directly

PR #200, opened 2016-10-25
[doc] adding debounce time unit

gasp:
it was not obvious


rt2zz:
:) imo all js times should be ms unless otherwise specified

PR #205, opened 2016-11-10
Attempt to set/get item from localstorage

### LINKED ISSUE #204: Access is Denied
I have started seeing `Access is Denied` runtime errors come in from our users.  I traced it down to [this line](https://github.com/rt2zz/redux-persist/blob/master/src/defaults/asyncLocalStorage.js#L26).  I believe the `hasLocalStorage` check should also attempt to set and get an item.

For more background on the problem, you can check out [this stackoverflow](http://stackoverflow.com/questions/13102116/access-denied-for-localstorage-in-ie10).

I'm willing to submit a PR if you're interested.

IanVS:
Fixes #204 

In Windows environments, it's possible for browsers to have a localStorage and yet not have access to it.  This ends up throwing an `Access is denied` error when attempting to get or set from localStorage.  

This change adds some additional safety checks into `hasLocalStorage` to set, get, and remove a test item inside the try block.

### DIFF
--- a/src/defaults/asyncLocalStorage.js
+++ b/src/defaults/asyncLocalStorage.js
@@ -9,9 +9,17 @@ const noStorage = process && process.env && process.env.NODE_ENV === 'production
   }
 
 function hasLocalStorage () {
+  let storageExists
   try {
-    return typeof window === 'object' && !!window.localStorage
+    storageExists = (typeof window === 'object' && !!window.localStorage)
+    if (storageExists) {
+      const testKey = 'redux-persist localStorage test'
+      window.localStorage.setItem(testKey, true)
+      window.localStorage.getItem(testKey)
+      window.localStorage.removeItem(testKey)
+    }
   } catch (e) { return false }
+  return storageExists
 }
 
 function hasSessionStorage () {


rt2zz:
interesting. Can we simplify the check to simply running a `localStorage.getItem('test key')` call? I would assume that would throw and there is no need to also test set and remove?

rt2zz:
also does sessionStorage suffer from the same quirk?

IanVS:
>Can we simplify the check to simply running a localStorage.getItem('test key')

That would handle this particular case, yes.  But, it does not save us if localStorage is at capacity and cannot be written to, which is where the `setItem` is useful.

>does sessionStorage suffer from the same quirk?

Honestly I don't know.  I don't think I've been seeing any sessionStorage errors, so I wouldn't know how to test it.  I'd say leave it alone unless it's determined to be a problem.

IanVS:
This is a pretty good thread with different options for checking localStorage: https://mathiasbynens.be/notes/localstorage-pattern#comment-9

rt2zz:
👍 I decided to simplify the check to just getItem for now. This module is very sensitive to startup time and I want to be conservative with anything that could impact this.

PR #207, opened 2016-11-12
add config.serialize as boolean

rt2zz:
**breaking change**

This is a proposed breaking change for v4 which will turn the (undocumented) serialize & deserialize config options from functions into a single serialize option as a boolean.

This simplifies the api, makes disabling serialization easier, and if pluggable serialization is needed it can (and should) be done through the transforms api.

### DIFF
--- a/redux-persist.js.flow
+++ b/redux-persist.js.flow
@@ -21,6 +21,7 @@ declare module 'redux-persist' {
     storage?: Storage,
     transforms?: Array<Object>,
     debounce?: number,
+    serialize?: boolean,
   }
   declare type Purge = (keys: Array<string>) => void
   declare type Rehydrate = (incoming: Object, options: { serial: boolean }) => void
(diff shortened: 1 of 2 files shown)


marvinhagemeister:
I'm all for simplifying code but I must admit that I haven't used them myself yet. The default json serializer may be enough for most people. I'm a bit afraid though when people try to put none serializeable objects in their store. Things like date objects, etc.

rt2zz:
@marvinhagemeister I agree, I always prefer simple serializable state. I will say however that I have seen a couple of large projects successfully rely on the transit serialization from [transform-immutable](https://github.com/rt2zz/redux-persist-transform-immutable) which handles Date objects well enough. Perhaps we should make that the suggested solution for people who want js dates in state.

Sounds like you agree custom serializer function is overkill 👍

marvinhagemeister:
@rt2zz That sounds great 👍  I must admit that I haven't really worked with immutablejs yet.

PR #211, opened 2016-11-15
Added note on the important ordering of transforms

swernerx:
This part was pretty time consuming for me a probably other users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,14 @@ let myTransform = createTransform(
 persistStore(store, {transforms: [myTransform]})
 ```
 
+Please keep care of the orders in which transform are applied e.g. 'compress' has to be used as the first transform in most cases like when used in combination with 'immutable':
+
+```js
+transforms: [ createCompressor(), immutableTransform() ]
+```
+
+Otherwise you might see JSON parse errors from deep inside of the [transit implementation](https://github.com/glenjamin/transit-immutable-js).
+
 ## Migrations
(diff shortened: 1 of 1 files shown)


rt2zz:
thank you for looking into this and the PR! However I am actually quite suprised, I would expect the required order to be `transforms: [ immutableTransform(), createCompressor() ]` since we [reduce left](https://github.com/rt2zz/redux-persist/blob/master/src/createPersistor.js#L57) the transforms and it would not make sense to run transit-immutable on top of something the is already a string (the output of createCompressor).

I have a suspicion that there is actually a bug in redux-persist-trasnform-compress which is causing this. I will have to follow up tomorrow.

rt2zz:
just published bugfix where compress was inconsistently applying serialization. Perhaps this fixes the issue for you?

swernerx:
What exactly should I test now? Putting immutable before compressor?

rt2zz:
yes. since the transforms apply left to right when storing state, it would be best to first immutable and then compress. In the other order I would not expect immutable to work, although please correct me if I am wrong.

swernerx:
Seems to work now correctly. Thanks.

Usage:

```js
      persistStore(this.props.store, {
        storage: localforage,

        // Order is important. Immutable has to be the first!
        transforms: [
          immutableTransform(),
          createCompressor()
        ],
        blacklist: [
          "apollo"
        ]
      }
```

rt2zz:
👍 thanks for bearing with and reporting!

PR #216, opened 2016-11-21
use a different immutable signature function

matttti:
There are immutable libs like https://github.com/rtfeldman/seamless-immutable which don't specify mergeDeep(). Instead, using asMutable() does not only sound more significant here, it is also part of facebook's immutable, and probably avoids unintended behavior with non-immutable objects implementing the generic name mergeDeep().


marvinhagemeister:
@rt2zz  LGTM

rt2zz:
indeed 🚀

Issue #120, opened 2016-06-15
Manual Persist


KarlPurk:
Is it possible to manually persist state rather than having state automatically persisted on each change?

rt2zz:
not currently, but I would be open to ideas for what this would look like

rt2zz:
@KarlPurk can you elaborate on why you want to manual trigger persist? If performance is an issue you can set a debounce on the persist, or also blacklist expensive reducers.

KarlPurk:
For development purposes.  Manually persisting state is useful when debugging issues where the application has to be in a particular state in order to reproduce the bug.  You can setup the application in the desired state and manually persist the state to storage - you're then in a position where you can make code changes and refresh the application to return to the same state quickly, allowing for easier debugging.

simonacca:
@KarlPurk Maybe you can try solving this problem with hot reloading.

A possible production use case for this feature is the following: I have set a rather long `debounce` for performance reasons and would like to be sure that the state is persisted when the user exits my app.

The ideal interface, I think, would be something like this:

```javascript
const persistor = persistStore(store, { debounce: 5000 });

window.onbeforeunload = function(){
  persistor.persist();
};
```

rt2zz:
@simonacca clever. The api is not exposed for this currently, but I am open to suggestions. `persistor.persist()` seems ambiguous (ie. it is *already* persisting). Some ideas:
`persistor.flush()`
`persistor.persistImmediate()`

This method would probably need to return a promise as persistence is typically (depending on storage engine) not sync.

simonacca:
Agreed, `.flush()` or `.persistImmediate()` is a much better API.

I am trying to implement this in #217 , let me know what you think.

fresk-nc:
it would be cool if I could persist the state even though it has not changed

wwwdata:
Another use case that we are currently having:

We have a native app which also has a react native app. So certain parts of the app are implemented with react native and are using redux-persist. Before shutting down the react native part of the app, we want to make sure all the redux state got persisted and we don't kill of the react native (in terms of Android) activity too early before the state could be persisted.

rt2zz:
`flush` is now part of the api, `persistor.flush()` which returns a promise that resolves when the flush is complete.

carlgieringer:
[...]

PR #217, opened 2016-11-21
Implements manual persist #120

simonacca:
I am implementing manual persist as of #120.

I got it working, but I don't understand the logic of the original code here: https://github.com/rt2zz/redux-persist/blob/master/src/createPersistor.js#L39-L63

Why adding to `storesToProcess` with every store update instead of just right before persisting ?
Why then process just one key at a time out of `storesToProcess` ?

Can we just wait for the timer to expire, then push all the stores in `storesToProcess` and finally loop over all the keys present in `storesToProcess` with the reduce function, as it is in the pull request ?

Cheers,
Simon

### DIFF
--- a/src/createPersistor.js
+++ b/src/createPersistor.js
@@ -32,44 +32,36 @@ export default function createPersistor (store, config) {
   let timeIterator = null
 
   store.subscribe(() => {
-    flush()
+    if (paused) return
+
+    // time iterator (read: debounce)
+    if (timeIterator === null) {
+      timeIterator = setInterval(() => {
+        flush()
+        timeIterator = null
+      }, debounce)
+    }
   })
-  
+
   function flush() {
-    return new Promise(
-      function(resolve, reject){
-        if (paused) resolve()
+    let state = store.getState()
 
-        let state = store.getState()
+    stateIterator(state, (subState, key) => {
+      if (!passWhitelistBlacklist(key)) return
+      if (stateGetter(lastState, key) === stateGetter(state, key)) return
+      storesToProcess.push(key)
+    })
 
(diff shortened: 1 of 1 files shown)


rt2zz:
@simonacca ty for PR. The time iterator is actually important for the performance of redux-persist. Essentially what is happening is we create a queue of stores that need processing and then pull out of that queue every so often (every config.debounce).

I think your PR has the right idea, but it actually needs to override a different level of the process. Something like: https://gist.github.com/rt2zz/950c4323a1e8b8eb9368aae973a325fa (untested, but I think close to where it needs to be)

simonacca:
Yes, I get how the code works, but I don't understand why it would be better performance-wise.

With the present implementation, you check which stores need processing every time `store.subscribe` produces an event.

Isn't it more efficient to limit ourselves to the setup of `timeIterator` within `store.subscribe`, to then check which substates of the stores need persisting just once, when the timer expires?

Suppose you set `debounce` to `1000`, then you update the store `100` times within `10 ms`:
with the code in the pr, the persistor would do nothing at all (except for checking whether a timer is set) for every `store.subscribe` call.
And finally check which substates to persist, and persist them just once, altogether, when the timer expires.

rt2zz:
building the queue is quite cheap (object equality on old state vs new state). We could break this up accordingly to the debounce as well, but in practice that makes negligible impact on performance.

All of the expensive work happens in serializing and writing the string to storage, so that is the portion we want to break up and avoid doing all at once (except in the case of flush)

rt2zz:
closing due to inactivity. Key issue here is making sure default perf is unchanged by breaking up the serilaization calls.

PR #219, opened 2016-11-22
Cleanup createTransform inbound test

kevmannn:
Noticed a race condition in the test for `createTransform inbound`.
Made use of `async / await` and `persistStore`'s callback to hopefully get around this.
_Edit:_ Realizing I misunderstood this - giving it another try.

### DIFF
--- a/test/createTransform.spec.js
+++ b/test/createTransform.spec.js
@@ -6,9 +6,8 @@ import { createTransform, getStoredState, persistStore } from '../src'
 import createMemoryStorage from './helpers/createMemoryStorage'
 const ACTION_MUTATE_ALL = 'ACTION_MUTATE_ALL'
 
-const createReducer = (actionCallback) => {
+const createReducer = () => {
   return (state = {a: 1, b: 2}, action) => {
-    actionCallback && actionCallback(action)
     if (action.type === ACTION_MUTATE_ALL) {
       return {a: 10, b: 20}
     }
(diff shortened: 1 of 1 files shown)


rt2zz:
👍 this is much better than those arbitrary set timeouts

You mentioned race condition, were you getting sporadic fails before?

kevmannn:
Yep - now the test is passing consistently on my machine (though not w/ travis, as you can see).
Do you think this is more or less the correct approach?

test/createTransform.spec.js
```diff
@@ -27,24 +26,26 @@ test('createTransform outbound', async function (t) {
...
-  let memoryStorage = createMemoryStorage(seedState)
-  persistStore(store, {storage: memoryStorage, transforms: [testTransform]})
 
-  setTimeout(() => {
+  const memoryStorage = createMemoryStorage(seedState)
+  const testTransform = createTransform(inbound, outbound)
+  const config = {storage: memoryStorage, transforms: [testTransform]}
+  persistStore(store, config, (err, restoredState) => {
+    t.ifError(err)
+    // t.deepEqual(restoredState, {a: 'hi', b: 'hi'})
```
kevmannn: @rt2zz This assertion would be a way of implicitly testing the outbound transform.
Provided I can get this test to pass, do you think it makes sense to remove the other test in this file?

The flow of expected inbound / outbound state seems slightly more explicitly shown this way (imo).

rt2zz:
@kevmannn apologize, this dropped off my radar. I am fully onboard with rewriting this poorly written test. Going to close this due to inactivity, but feel free to reopen

PR #234, opened 2016-12-14
added index.d.ts for typescript

morlay:


package.json
```diff
@@ -39,6 +40,7 @@
   "author": "rt2zz <zack@root-two.com>",
   "license": "MIT",
   "dependencies": {
+    "@types/redux": "^3.6.0",
```
rt2zz: I am not familiar with typescript, how does this dep work?

rt2zz:
awesome! looking forward to merging this. See my question about how that dep works though ^

package.json
rt2zz: I am concerned about the bytes that the redux typings will add. Shouldnt the typings field be sufficient and we can may @types/redux a devDependency for the sake of development?

morlay: It will not, there are no js file under `@types/xxx`.
However, it is fine to move it to devDeps, for end user, `@types/redux` will be added with `redux`.

rt2zz:
thank you! I went ahead and moved it to devDeps just because that feels more comfortable

morlay:
Cool. but forgot to put it in `files` of package.json

marvinhagemeister:
@morlay Awesome stuff! I've been meaning to add typings for a while now, but haven't had the time to do so. That's what I love about open source 👍  Thanks for contributing 💯

rt2zz:
@morlay I just updated the location and build step for types: https://github.com/rt2zz/redux-persist/commit/ff42d2f00c4148b9daf1669b22ec64a575731d4c

Do you mind giving that a look over and make sure it still works for the ts case?

PR #241, opened 2016-12-21
Improve typescript definitions

### LINKED ISSUE #240: Typings to typescript are incomplete
Typings are missing `keyPrefix` property in config. Also some types in case of transform are not really necessary. There is no `interfaces` module in original library so this shouldn't appear only in typings. 

@morlay You created typings in "old" style. Without knowing you were working on these I also created typing for this library and it has been already merged into `@types` in npm. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/13389
Maybe we should merge yours with mine for better support? Who should take care of this?

LKay:
Resolves #240

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,8 +9,8 @@
   "scripts": {
     "test": "standard 'src/**/*.js' 'test/**/*.js' storages.js constants.js && BABEL_ENV=commonjs ava",
     "test:watch": "npm test -- --watch",
-    "copy-types": "cp -r type-definitions/ lib/",
-    "clean": "rimraf lib dist es",
+    "copy-types": "rsync -av --exclude=constants.d.ts --exclude=storages.d.ts type-definitions/ lib/ && cp type-definitions/constants.d.ts type-definitions/storages.d.ts ./",
+    "clean": "rimraf lib dist es && rm -f constants.d.ts storages.d.ts",
     "build": "npm run build:commonjs && npm run build:es && npm run build:umd && npm run build:umd:min && npm run copy-types",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
     "build:es": "cross-env BABEL_ENV=es babel src --out-dir es",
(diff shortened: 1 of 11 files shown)


rt2zz:
awesome, @morlay do you mind giving this a look over as well?

package.json
```diff
@@ -9,8 +9,8 @@
   "scripts": {
     "test": "standard 'src/**/*.js' 'test/**/*.js' storages.js constants.js && BABEL_ENV=commonjs ava",
     "test:watch": "npm test -- --watch",
-    "copy-types": "cp -r type-definitions/ lib/",
-    "clean": "rimraf lib dist es",
+    "copy-types": "rsync -av --exclude=constants.d.ts --exclude=storages.d.ts type-definitions/ lib/ && cp type-definitions/constants.d.ts type-definitions/storages.d.ts ./",
```
rt2zz: having to colocate type files is a messy unforeseen consequence of putting storages and constants in root :(

In particular for storages we do it to help with dead code elimination for projects that do not use the default storage.

is there no other way to tell typescript where to look for these definitions?

LKay: Not really, I'm afraid, if you want to be fully compatible with typescript 2.x definitions. It is possible to use it now as single bundle file of the structure that @morlay created but it is now discouraged and likely will be deprecated in some future versions of typescript.

morlay: I am not sure the module declaration will be deprecated, at lease they will be useful for patching, when some issues in origin one. ([example](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-plugin-d-ts.html), working by the [declaration-merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) rule)

morlay: BTW. I prefer to keep module declaration and with in single one file, it is more friendly for project which is not wrote by TypeScript (TypeScript could auto generate declarations by compiler).

rt2zz:
@morlay @LKay not 100% sure best path forward here (not having used ts myself). If we can get away with a single .d.ts file I would greatly prefer that. Let me know if I can do anything here to help.

LKay:
Sorry for late reply, I missed the notifications for the repo completely. Since the version of the library updated and definitions have changed significantly I will close this. I will need to make few fixes to the existing ones definitions but will address these in separate PR.

PR #245, opened 2017-01-03
autoRehydrate() does object comparison correctly using JSON.stringify

jeffreycrow:
Right now autoRehydrate uses a comparison check of `state[key] !== reducedState[key]` to see if the reducer changes the substate. The !== operator will always return true when comparing two objects.

This implementation converts the objects to strings using JSON.stringify and then does the comparison, which allows for a true deep comparison of the objects just based on content.

This may not be the best way to do object comparison but it worked for me.

### DIFF
--- a/src/autoRehydrate.js
+++ b/src/autoRehydrate.js
@@ -58,7 +58,7 @@ function defaultStateReconciler (state, inboundState, reducedState, log) {
     }
 
     // if reducer modifies substate, skip auto rehydration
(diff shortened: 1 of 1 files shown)


rt2zz:
unfortunately JSON.stringify is expensive. We assume immutability and thus do reference equality as most redux reducers are immutable (if you need to change a value, make it a new object). 

If your application specifically needs it to behave this way, I would recommend making a custom auto rehydrate that you store and consume locally which has this behavior.

PR #251, opened 2017-01-10
Update README with usage example and explanation

### LINKED ISSUE #246: README should show usage with `compose` and `applyMiddleware`
For many people, redux-persist is the first library they use which exposes an enhancer, rather than a middleware. This seems to cause a lot of confusion for people who don't read much past the README. Currently the opening example is:

```js
import {persistStore, autoRehydrate} from 'redux-persist'
const store = createStore(reducer, undefined, autoRehydrate())
persistStore(store)
```

Maybe it'd be friendlier to show a more realistic example, since everyone uses `applyMiddleware` in real life:

```js
import {compose} from 'redux'
import {persistStore, autoRehydrate} from 'redux-persist'
const store = createStore(reducer, undefined, compose(applyMiddleware({}), autoRehydrate())
persistStore(store)
```

I'd happily open a PR to do so, but I thought I should ask first.

ucarion:
Closes #246.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,10 +10,22 @@ Redux Persist is [performant](#why-redux-persist), easy to [implement](#basic-us
 [![npm downloads](https://img.shields.io/npm/dm/redux-persist.svg?style=flat-square)](https://www.npmjs.com/package/redux-persist)
 
 ## Basic Usage
-Basic usage requires adding three lines to a traditional redux application:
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
@ucarion Thanks a lot for your PR. Sorry, for not merging this sooner 🎉

PR #255, opened 2017-01-17
Return Promise from purge() in persistStore.js

### LINKED ISSUE #254: purge() in persistStorage.js should return a promise
Hi, I noticed that the `purge()` function in [persistStorage.js](https://github.com/rt2zz/redux-persist/blob/master/src/persistStore.js) doesn't return the promise that the `persistor.purge()` returns. 

I'll open a PR with a fix.

domlazic:
Fixes #254

### DIFF
--- a/src/persistStore.js
+++ b/src/persistStore.js
@@ -42,7 +42,7 @@ export default function persistStore (store, config = {}, onComplete) {
     ...persistor,
     purge: (keys) => {
       purgeKeys = keys || '*'
-      persistor.purge(keys)
+      return persistor.purge(keys)
     }
   }
 }


rt2zz:
ty

domlazic:
np, glad to help. Do you maybe have an idea when will you update the npm package to include this change? thanks!

rt2zz:
done https://github.com/rt2zz/redux-persist/commit/676e68b930057f62cb4e7501ae3d764386a41928

SleeplessByte:
@rt2zz cc @domlazic type definitions are not updated to reflect this change.

rt2zz:
@SleeplessByte indeed good catch. PR greatly appreciated!

PR #258, opened 2017-01-23
Remove key check for autoRehydrate.js

### LINKED ISSUE #130: autoRehydrate does not lift the reducer it makes
Finally found the root cause...

Problem:
if you use redux-devtools & call `replaceReducer`, the initial state overrides the persisted state.

Reason:
`replaceReducer` doesn't know about the reducerEnhancer that `autoRehydrate()` creates. That means the devtools inits without it, and doesn't merge the initial state with the persisted state.

Reproduction:
Call `store.replaceReducer` on an async route. (Probably works on sync routes, too. Haven't tried).

Workarounds:
None (use redux-storage?)

Fix:
Either offer a reducerEnhancer like redux-storage or lift the `replaceReducer` into the `liftedStore`. I think the former is cleaner because `autoRehydrate` does not need a lifted store, so using a higher-level API seems reasonable.

jnishiyama:
In order to resolve #130 **if** it is unintended that asynchronously loaded reducers will not be hydrated by `autoRehydrate` store enhancer. Removes the check for persisted keys to exist in Redux initial state.

### DIFF
--- a/src/autoRehydrate.js
+++ b/src/autoRehydrate.js
@@ -48,8 +48,6 @@ function defaultStateReconciler (state, inboundState, reducedState, log) {
   let newState = {...reducedState}
 
   Object.keys(inboundState).forEach((key) => {
-    // if initialState does not have key, skip auto rehydration
-    if (!state.hasOwnProperty(key)) return
 
     // if initial state is an object but inbound state is null/undefined, skip
     if (typeof state[key] === 'object' && !inboundState[key]) {
--- a/src/autoRehydrate.js
+++ b/src/autoRehydrate.js
@@ -48,7 +48,6 @@ function defaultStateReconciler (state, inboundState, reducedState, log) {
   let newState = {...reducedState}
 
   Object.keys(inboundState).forEach((key) => {
-
     // if initial state is an object but inbound state is null/undefined, skip
     if (typeof state[key] === 'object' && !inboundState[key]) {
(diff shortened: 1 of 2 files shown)


rt2zz:
@jnishiyama sorry for the delay in taking a look. Yes I think you bring up a very valid shortcoming of the current approach. The upside is that bad state is less likely to stick around. 

I am working on a new form of autoRehydrate https://github.com/rt2zz/redux-persist-state-manager which among other things I has support for migrations built in which makes me more comfortable being more aggressive in taking persisted data.

Its early, but if you are open to switching, I would gladly accept this PR over there. I will try to put up a readme this week.

rt2zz:
note: my hesitance to merging this is because this project is used by many people and this would technically be a breaking change.

rt2zz:
@jnishiyama I just updated redux-persist-state-manager with this change. It now only logs a warning (when `log: true`) but does not skip the value. I also updated the docs. Let me know if that works for your case.

https://github.com/rt2zz/redux-persist-state-manager/blob/master/src/index.js#L72

jnishiyama:
@rt2zz Hey, it should work for my case, is there a large difference between the new form of AutoRehydrate and the old form?

rt2zz:
@jnishiyama it is a higher order reducer, and hence should play nicer with hot reloading.

Main breaking change is that it hard sets incoming state instead of shallow merging.

rt2zz:
I think the recommendation here is to go with redux-persist-state-manager

We will make that approach the default in v5, but I think that is still a ways off 👍

PR #275, opened 2017-02-10
[docs] purgeAll -> purge

zalmoxisus:
Seems like `purgeAll` is not present in the API and `purge` without argument purges all keys.


rt2zz:
Ah thank you much! purgeAll was deprecated in v3 and removed in v4

PR #286, opened 2017-02-21
Fixed TypeScript error

joanrodriguez:
Typescript throws errors when parsing now that you have set `getAllKeys` as optional... 
Both in webpack and VisualStudio Code.

### DIFF
--- a/type-definitions/index.d.ts
+++ b/type-definitions/index.d.ts
@@ -5,7 +5,7 @@ declare module "redux-persist" {
     setItem(key: string, value: any, onComplete?: OnComplete<any>): Promise<any>;
     getItem<Result>(key: string, onComplete?: OnComplete<Result>): Promise<Result>;
     removeItem(key: string, onComplete?: OnComplete<any>): Promise<any>;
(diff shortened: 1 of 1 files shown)


marvinhagemeister:
Thanks for the fix. Sorry that this issue got through.

rt2zz:
@joanrodriguez getAllKeys needs to be optional in order for localforage to type. perhaps the correct syntax is `getAllKeys?<Result>(onComplete?: OnComplete<Result>): Promise<Result>;`?

marvinhagemeister:
@rt2zz Sorry for merging this too quickly. The previous typings in master where invalid and wouldn't compile at all. Just looked at the source and you're right that `getAllKeys` needs to be optional. In that case the correct typings would be:

`getAllKeys?: <Result>(onComplete?: OnComplete<Result>) => Promise<Result>;`

Nonetheless we really need to add tests for our typings. There are further errors with the `Storage` and the get/setItem signature.

rt2zz:
:(

Agreed about adding typings tests. I am really wanting typings to come from the community, specifically users of said type systems. Perhaps we make type tests a requirement for the next typings PR to come through.

Issue #263, opened 2017-01-25
Cant find variable: Reflect, appears after upgrading RN 0.40


wachunei:
After upgrading to RN 0.40 I keep getting this error when purging some keys at rehydration.
It happens only on device and not in simulator, which is strange.

redux-persist@4.1.0
react-native@0.40.0

Any ideas?

rt2zz:
I have not seen this, and have been running redux-persist 4.x against rn 0.40 for some time now.

If this continues to be an issue I am open to removing the use of reflect here: https://github.com/rt2zz/redux-persist/blob/1cd5d29ab49b96be1026e8bab67aaeae1699e9d9/src/persistStore.js#L27

wachunei:
@rt2zz It has still been happening. I'm guessing that API may be not available in every production environment and we can't rely on expecting that to be true, maybe can we use a polyfill as a dependency?

rt2zz:
@wachunei perhaps, but I doubt the benefits of Reflect.delete are worth the costs of polyfilling. `delete` should work just as well in this case.

marcinkornek:
The same happen for me with `react-native 0.42.0`. When I tried to remove all keys with `purge()` it did nothing (it didn't delete items) but when I tried to purge array of keys it showed me `Cant find variable: Reflect` error.

wachunei:
closed by #291

PR #291, opened 2017-03-04
Remove use of Reflect to delete key

wachunei:
This removes use of Reflect API to delete persistedState object key, since it's been reporting error under certain evironments. See #263


rt2zz:
thank you very much!

PR #295, opened 2017-03-14
check if bound is a function

stevemao:
So if we don't need to define our custom inbound, we can pass null.
```js
createTransform(
  null,
  (state) => (state)
)
```

### DIFF
--- a/src/createTransform.js
+++ b/src/createTransform.js
@@ -9,8 +9,8 @@ function createTransform (inbound, outbound, config = {}) {
   }
 
   return {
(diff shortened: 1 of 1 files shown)


rt2zz:
null will go to the second side of the ternary here: `inbound ? inbound(state, key) : state` so this functionality should be unaffected by your PR. 

I see nothing wrong with the function check, but is it needed? we have TypeScript and flow type support if it is an underlying typing issue.

rt2zz:
closing due to inactivity, let me know if you think this is still necessary

PR #304, opened 2017-03-22
(webpack): update config for newly required absolute paths

rt2zz:


rt2zz:
appears to have just started: https://travis-ci.org/rt2zz/redux-persist/jobs/213698277

if only we had yarned

PR #308, opened 2017-03-27
Export consts as string literals

unindented:
Instead of exporting them as generic `string`s, I propose they are exported as string literals (https://www.typescriptlang.org/docs/handbook/advanced-types.html#string-literal-types). That way users can do discriminated unions (https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions).


rt2zz:
Good call 👍

Issue #305, opened 2017-03-22
Dead code elimination problem


Andarist:
This https://github.com/rt2zz/redux-persist/blob/2bbc25a9dee2ca54f521632c17810642c1352fa1/src/env.js#L1 should be avoided. 

Why? Uglify wont be able to eliminate this if we store it as variable. Basically what is done to `process.env.NODE_ENV` with webpack is that its geting replaced in the source code (think about strings) with the value, so later on uglify sees something like `if ('development !== 'development')` which can be replaced to `if (false)` and this branch of code can be safely excluded from the dist file.

Also referencing `process` alone, causes webpack to polyfill it automatically, which doesnt happen when using it like this `process.env.NODE_ENV` due to string replacement thing ive described.

I can provide a PR to fix this, just let me know.

rt2zz:
Thank you for pointing this out. Yes PR would be greatly appreciated, IIRC we recently switched to the process check because users without webpack were getting a process is undefined error. 

What is the best way to restore this functionality while still shipping a working UMD build?

Andarist:
Hm, yeah I've tried to do exactly the same in redux-saga/redux-saga for us, as the same thing got reported but while guarding I've noticed that [this](https://github.com/redux-saga/redux-saga/blob/ab5cd83f8f006c02136061afcbb91b0d270e5bbb/dist/redux-saga.js#L2354-L2532) was added to the dist file, just because `process` alone was referenced in the code.

Truth to be told I have no idea what to do about this issue for non-webpack/rollup users. You can also check redux's issue about kinda same thing - https://github.com/reactjs/redux/issues/2244 , where `process.env.NODE_ENV` was cached for a brief moment - not only adding `process` polyfill but also preventing dead code elimination, and they had to revert the change.

It's also totally suprising that such, one would think, irrelevant change causes all of this. This auto-behaviour should be documented better.

What kind of distribution caused problems for u? In general I think just distributing dist files shouldnt be a problem as you are the creator of those and therefore using a webpack to do the work. For me its just other bundler (Aurelia's one) which is causing a problem as users are trying to compile source code by themselves.

rt2zz:
Great detailed post! This is fantastic - and yes this is way to easy to break! 

Feel free to PR, else I will likely get this in myself this weekend.

Andarist:
I could provide a PR, but probably not in this week as Im leaving for the weekend and I guess I wont have too much time. 

[...]

PR #309, opened 2017-03-28
Update NODE_ENV checks

rt2zz:
follow up to https://github.com/rt2zz/redux-persist/issues/305

process.env.NODE_ENV is better for webpack/uglify and moreover is the standard in react/redux


marvinhagemeister:
Awesome!

PR #310, opened 2017-03-28
(storages): remove nextTick in favor of setImmediate

rt2zz:
reference #305


marvinhagemeister:
Yay, for smaller bundles 👍

PR #314, opened 2017-03-30
Add Prettier

rt2zz:
lets get on this bandwagon ;)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "build:umd": "cross-env BABEL_ENV=commonjs webpack",
     "build:umd:min": "cross-env BABEL_ENV=commonjs NODE_ENV=production webpack",
     "build:watch": "npm run build:commonjs -- -watch",
+    "precommit": "lint-staged",
     "prepublish": "npm run clean && npm run build"
   },
   "lint-staged": {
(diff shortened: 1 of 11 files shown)


marvinhagemeister:
Awesome! Prettier is a relief for the js community 💯

package.json
```diff
@@ -17,8 +17,16 @@
     "build:umd": "cross-env BABEL_ENV=commonjs webpack",
     "build:umd:min": "cross-env BABEL_ENV=commonjs NODE_ENV=production webpack",
     "build:watch": "npm run build:commonjs -- -watch",
+    "precommit": "lint-staged",
     "prepublish": "npm run clean && npm run build"
   },
+  "lint-staged": {
```
randycoulman: I love Prettier, and using lint-staged here is super handy for contributors.  But be aware that this recipe can be very surprising if someone has used `git add -p`: any file that gets reformatted will become part of the commit, even if it was only partially staged.  See https://github.com/okonet/lint-staged/pull/75 for details.

rt2zz: do you have an alternate recommendation for how to implement prettier? I would like to ensure all code PRed is prettier'ified, wether that is via CI or lint-staged or other.

randycoulman: For now, I'm living with doing `prettier --list-different` in my pre-commit hook to fail the commit on unformatted files and then adding a `format` script to actually do the formatting. Not ideal, but better than surprising people IMO.

rt2zz:
this is on v5 branch

Issue #311, opened 2017-03-28
customizing storage key.


vaukalak:
In certain apps you need a portion of data to be fragmented in base of action user, or share among different users.

I think the easiest solution would be to pass within `config` parameter the `createStorageKey` with current `createStorageKey` as default implementation (probably named as `defaultCreateStorageKey`).

Also I think interface of that function should be changed to:

```
function createStorageKey (keyPrefix, key, state) {
    return `${keyPrefix}${key}`
 }
```

this will allow us to do things like:

```
function createStorageKey(keyPrefix, key, state) {
    let myKey = key;
    if (key === 'somePrivateInfo') {
      myKey = `${state.user.id}-${key}`;
    }
    return `${keyPrefix}${myKey}`
 }
```

rt2zz:
interesting idea. A change like this might not be so trivial, for example when rehydrating, before we have grabbed the user state, how do we know which keys to rehydrate?

vaukalak:
@rt2zz I think, that developer that uses that feature, will need to load data, he'll load just a part of the state. and yep, no persist actions to fragmented reducers should before user logins, or logout (which is logical). But I think, all that stuff should be considered on the application level.

vaukalak:
Sorry, I'll rephrase. I mean it's a task of developer, who will use the lib to take care, that there no writes to reducer fragmented by user, and to load those reducers later.

vaukalak:
@rt2zz did you have a chance to have a thought on that?

Xuhao:
Facing same problem here.
In our chat app, we hope user data separated from each other, it's like fill state with history data after user logged in, maybe set user.id as prefix/suffix of the storage key dynamically.

PR #320, opened 2017-04-03
#311 implemented fragmented keys

vaukalak:
this PR solves #311

### DIFF
--- a/src/createPersistor.js
+++ b/src/createPersistor.js
@@ -12,6 +12,7 @@ export default function createPersistor (store, config) {
   const transforms = config.transforms || []
   const debounce = config.debounce || false
   const keyPrefix = config.keyPrefix !== undefined ? config.keyPrefix : KEY_PREFIX
+  const createFragmentedKey = config.createFragmentedKey || defaultCreateFragmentedKey
 
   // pluggable state shape (e.g. immutablejs)
   const stateInit = config._stateInit || {}
@@ -53,8 +54,16 @@ export default function createPersistor (store, config) {
         }
 
         let key = storesToProcess.shift()
-        let storageKey = createStorageKey(key)
-        let endState = transforms.reduce((subState, transformer) => transformer.in(subState, key), stateGetter(store.getState(), key))
+        let storageKey = createStorageKey(keyPrefix, key, state)
+        // storageKey can be undefined, if there's no info to generate it
+        if (!storageKey) {
+          storesToProcess.shift()
+        }
+        let endState = transforms.reduce(
+          (subState, transformer) =>
+            transformer.in(subState, key),
+            stateGetter(store.getState(), key)
+        )
         if (typeof endState !== 'undefined') storage.setItem(storageKey, serializer(endState), warnIfSetError(key))
       }, debounce)
     }
@@ -88,8 +97,12 @@ export default function createPersistor (store, config) {
     return state
   }
 
-  function createStorageKey (key) {
-    return `${keyPrefix}${key}`
+  function defaultCreateFragmentedKey (key) {
+    return key
+  }
+
+  function createStorageKey (keyPrefix, key, state) {
+    return `${keyPrefix}${createFragmentedKey(key, state)}`
   }
 
   // return `persistor`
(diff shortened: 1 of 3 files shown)


vaukalak:
@rt2zz is there any chance, this will be reviewed?

src/createPersistor.js
```diff
@@ -12,6 +12,7 @@ export default function createPersistor (store, config) {
   const transforms = config.transforms || []
   const debounce = config.debounce || false
   const keyPrefix = config.keyPrefix !== undefined ? config.keyPrefix : KEY_PREFIX
+  const createFragmentedKey = config._createFragmentedKey || defaultCreateFragmentedKey
```
Andarist: if u pass some option in the config, it for sure shouldnt be named like this `_createFragmentedKey` which indicates something private

vaukalak:
@Andarist updated removed `_` in config keys?

vaukalak:
@rt2zz @Andarist sorry for bumping, but do you have any questions / notes on that?

src/getStoredState.js
```diff
@@ -55,14 +60,26 @@ export default function getStoredState (config, onComplete) {
...
-    if (whitelist && whitelist.indexOf(key) === -1) return false
-    if (blacklist.indexOf(key) !== -1) return false
+    if (whitelist && whitelist.map(mapKey).indexOf(key) === -1) return false
+    if (blacklist.map(mapKey).indexOf(key) !== -1) return false
     return true
   }
 
-  function createStorageKey (key) {
-    return `${keyPrefix}${key}`
+  function defaultFragmentKeyToReducerKey (key) {
```
Andarist: `defaultFragmentKeyToReducerKey` and `defaultCreateFragmentedKey` are the same and also they are just `identity` function 

`const identity = x => x`

Andarist:
Ive added a single comment, but truth to be told im not that much involved in the project and Im not making any calls on new features etc, cannot review this properly as Im lacking the context for the feature and the actual codebase.

rt2zz:
Apologize for the long delay, I have been working on getting v5 ready to go, and it is just about ready for primetime. I think v5 *may* solve the issue out of the box, or possibly some small modifications are required, lets discuss:

Please note, in v5 only **one** key/value is stored per persistReducer, so the solution is not directly analogous to the one in this PR.

If I understand the problem fully, the the basic goal is to have some portion of state partitioned based on some variable, in this case lets say its the `user` reducer, partitioned by `userId`. in v5 this can be modeled by only adding the `user` reducer once we know what the partition key is. so e.g.

```js
// event: user logs in
const persistConfig = { key: `user:${userId}`, storage }
const persistedUserReducer = persistReducer(persistConfig, userReducer)

store.replaceReducer(combineReducers({
  ...baseReducers,
  user: persistedUserReducer,
})
persistor.persist() // @NOTE this method does not yet exist.
```

If we want to go with this we need to add the persist method to the persistor which I have been thinking about doing regardless. 

Ergonomically I think the challenge here is that the partition key need to come from presumably other state. Which means we would always need to first persist the primary reducer, then persist again for the partitioned reducer. I do not see a way around this, but we might be able to make the external api simpler, e.g. avoid store.replaceReducer. That might look something like:
```js
let userReducer = persistReducer({ key, storage, persistKey: PERSIST_USER})

// ...

let persistor = persistStore(store, undefined, () => {
  let userId = store.getState().userId
  if (userId) {
    userReducer.replacePersistConfig({ key: `user:${userId}`, storage })
    persistor.persist(PERSIST_USER)
  }
})
```

I am not sure which approach I like more, the idea of adding a `replacePersistConfig` method to the reducer is not particularly appealing.

Thoughts?

ScreamZ:
Any news on that people ? Now we are on V5 #637 along other need something like that ?

vaukalak:
@ScreamZ I think this PR is good to close. @rt2zz can you please post duplicate your post in #637 . So we can discuss the desired api? thanks.

Faolain:
@rt2zz any shot you can duplicate your post on #637 as vaukalak mentioned?

PR #321, opened 2017-04-04
Expose an imperative flush() method on persistors

nevir:
Handy for when you're debouncing, but still want to ensure that state is written out.

Prior art:

* https://github.com/rt2zz/redux-persist/pull/217
* https://github.com/rt2zz/redux-persist/issues/120


src/createPersistor.js
```diff
@@ -27,6 +27,7 @@ export default function createPersistor (store, config) {
 
   // initialize stateful values
   let lastState = stateInit
+  let lastWritten = {} // Last written value, per key
```
rt2zz: why do we need to store lastWritten? that is more or less what lastState tries to track.

src/createPersistor.js
```diff
@@ -52,17 +51,47 @@ export default function createPersistor (store, config) {
...
+  function persistCurrentStateForKey (state, key) {
+    let storageKey = createStorageKey(key)
+    let currentState = stateGetter(state, key)
+    lastWritten[key] = currentState
+    let endState = transforms.reduce((subState, transformer) => transformer.in(subState, key), currentState)
+    if (typeof endState === 'undefined') return null
+    return storage.setItem(storageKey, serializer(endState), warnIfSetError(key))
+  }
+
+  function persistCurrentState () {
```
rt2zz: would be more performant to just write the existing keys that are pending (everything else is gauranteed to be up to date)

i.e.
```js
keys.forEach((key) => {
  persistCurrentStateForKey(state, key)
})
keys = []
```

nevir:
PTAL

src/createPersistor.js
```diff
@@ -52,17 +50,41 @@ export default function createPersistor (store, config) {
...
+    let state = store.getState()
+    let promises = storesToProcess.map(k => persistCurrentStateForKey(state, k))
+    storesToProcess.splice(0)
+    return promises
```
vovkasm: 1. `promises` can contain `null` values because `persistCurrentStateForKey` can return nulls. Is it desired behaviour? What if user will do `Primise.all(persistor.flush()).then(...)`?

2. May be better api for user will be to return single `Promise` object?

May be do something like:
```js
const promises = storesToProcess.map(k => persistCurrentStateForKey(state, k))
return Promise.all(promises.filter(v => v != null))
```

src/createPersistor.js
```diff
@@ -53,15 +51,40 @@ export default function createPersistor (store, config) {
...
+      if (stateGetter(lastState, key) === newState) return
+      keys.push(key)
+    })
+
+    return keys
+  }
+
+  function persistCurrentStateForKey (state, key) {
+    let storageKey = createStorageKey(key)
+    let currentState = stateGetter(state, key)
```
scotchapman: I had issues with `let currentState = stateGetter(state, key)` getting a stale state.

This worked for me: `let currentState = stateGetter(store.getState(), key)`

I'm not sure if it was because of this or something else:

> The listener should not expect to see all state changes, as the state might have been updated multiple times during a nested dispatch() before the listener is called. It is, however, guaranteed that all subscribers registered before the dispatch() started will be called with the latest state by the time it exits.

http://redux.js.org/docs/api/Store.html#subscribe

reisel:
hey.
any updates or estimation when it's going to be merged?
are there any issues you need help with? I'll be happy to take a look.

iamwebdesigner:
@rt2zz please merge this one, if this is a correct solution

doniyor2109:
+1

drhorse:
Can we merge? I really need this feature. Thx

drhorse:
When user changes app display language, i can use this flush() and restart the app after it is done. As a short-term solution, I now write this language option separately to AsyncStorage, and restart the app. When app restarts, if this separate item exists, I user reducer again to update the store...

pgambling:
Any chance this will be merged soon?

rt2zz:
woo, ok sorry it took half a year to finish this off. Thank you @nevir for your patience, I apologize for taking so long to getting around to it. 

In order to reduce the maintenance burden I applied this against v5, and is now included in `5.0.0-beta.13`. The api is `await persistor.flush()` which will trigger *all* persisted reducers to flush their queue and immediately write to disk.

rt2zz:
going to move forward with this as a v5 feature, 5.0.0 will be out next week.

PR #324, opened 2017-04-06
add function and regexp support to blacklist/whitelist of persister config

ronnin:
then it will supported:

`
persistStore(store, {
    blacklist: [
      /^_/,
      k => k.length > 20,
    ],
  }, onComplete)
`

states with name underscore(_) prefixed or longer then 20, will not be persisted


rt2zz:
@ronnin why did you close? I have been busy and did not have a chance to late a look at this. Would it make sense to instead allow blacklist/whitelist to be a function, which would make it trivial to allow any match pattern including regex?

Andarist:
+1 for function over regexp

vincent-louis:
was hoping to have this as well, to filter dynamic sensitive keys in reducers.

flavinger:
Agree, would still love to have regex support in whitelist & blacklist. Would it be possible to reopen?

PR #327, opened 2017-04-10
Add redux-persist-filesystem-storage as available storage engine

robwalkerco:
Due to the limitation on React Native Android of the amount of data that can be stored in a single persisted reducer, I have created an alternative storage implementation that uses the filesystem to persist the data rather than AsyncStorage.


rt2zz:
🎈

Issue #329, opened 2017-04-17
Open to improvements around setimmediate?


toranb:
I'm new to the project so please forgive me if some history has shown why we use `global.setImmediate` to pull in what I would otherwise assume comes from `npm install setimmediate`

https://github.com/rt2zz/redux-persist/blob/master/src/defaults/asyncLocalStorage.js#L1

Is it possible we could use an ES6 import here or will this always be a global lookup (when undefined)?

A little background: I'm currently working to add `redux-offline` and thus `redux-persist` support for emberJS developers and the very last blocker is this line so I'm curious what I can do to learn more about it /and possibly unblock it by using an ES6 import of some kind (if possible).

Thank you in advance!

rt2zz:
What is the problem with calling `global.setImmediate`?

IIRC this was the safest way to retrieve setImmediate from non-browser environments like react-native. It is also found in other places like this polyfill: https://github.com/davidmarkclements/setImmediate-min/blob/master/index.js

I am open to resolving this, but do not fully understand the issue. Let me know! I would prefer to not introduce a new dep.

toranb:
Fair question :) 

I'd like to improve this for browser only dev (like I'm facing today w/ emberJS). I've worked around it by doing a simple patch but wanted to see if we could ES6-ify this import to improve the experience for non react devs (assuming that is an option?). 

```js
var global = {
    setImmediate: window.setTimeout
  }
```

I'm open to be educated here - just curious how I can go about using this w/out hacks like you see above ^^

rt2zz:
I do not know - if anyone else can weight in that would be great.

It occurs to me however, there may be a much simpler solution, we can remove setImmediate calls from all of the default localStorage implementation and we can replaced the use in `persistStore` as folllows
`const genericSetImmediate = typeof setImmediate === 'undefined' ? global.setImmediate || function (fn) { return setTimeout(fn, 0) } : setImmediate`
->
`const genericSetImmediate = typeof setImmediate === 'undefined' ? (fn) => setTimeout(fn, 0) : setImmediate`

toranb:
@rt2zz I've been able to work around this (to avoid the global hack mentioned above). So for now I'd say "ignore me" and I'll do my best to learn more about react native/ other (non ember) envs that the community is using this great library :)

If you still want to hash out the change above I'm all ears - curious if others would find that useful

rt2zz:
great, I will still mull over these potential changes but maybe with a little less urgency :D

[...]

PR #334, opened 2017-04-25
Do not mention setImmediate in the code alone, so its polyfill doesnt…

### LINKED ISSUE #333: Polyfills included in builds
I've noticed that i.e. [this](https://github.com/rt2zz/redux-persist/blob/c3a9332c5218f663047d78e0206ca78b6c316c55/src/defaults/asyncLocalStorage.js#L1) and [this](https://github.com/rt2zz/redux-persist/blob/c3a9332c5218f663047d78e0206ca78b6c316c55/src/persistStore.js#L6) still requires some polyfills to be included automatically when bundled.

Both `global` (which is not a problem I think, it's really small) is required and `setImmediate` too.

Maybe we could change both to something like:
```javascript
const genericSetImmediate = typeof setImmediate !== 'undefined' ? global.setImmediate : fn => setTimeout(fn, 0)
```
This should produce a smaller bundle (I think, but needs to be tested) due to lack of requiring `setImmediate` which has also a dependency on `process`, so it requires it too.

WDYT?

Andarist:
… get included in the bundle (fixes #333 and #329)

@rt2zz Please review if it satisfies library's requirements.

By omitting direct mention of `setImmediate` library went down a little bit in terms of size:
```
redux-persist.js - 46K -> 33K
redux-persist.min.js - 16K -> 12K
```

Still would like to somehow omit also `global` polyfill, as its unneccessary too, but have no idea how to omit both at once and still provide expected functionality. At least `global`'s is pretty small.

### DIFF
--- a/src/defaults/asyncLocalStorage.js
+++ b/src/defaults/asyncLocalStorage.js
@@ -1,4 +1,4 @@
-const localSetImmediate = typeof setImmediate === 'undefined' ? global.setImmediate : setImmediate
+import setImmediate from '../utils/setImmediate'
 
 let noStorage = () => { /* noop */ return null }
 if (process.env.NODE_ENV !== 'production') {
@@ -57,7 +57,7 @@ export default function (type, config) {
           for (var i = 0; i < storage.length; i++) {
             keys.push(storage.key(i))
           }
-          localSetImmediate(() => {
+          setImmediate(() => {
             cb && cb(null, keys)
             resolve(keys)
           })
@@ -71,7 +71,7 @@ export default function (type, config) {
       return new Promise((resolve, reject) => {
         try {
           var s = storage.getItem(key)
-          localSetImmediate(() => {
+          setImmediate(() => {
             cb && cb(null, s)
             resolve(s)
           })
@@ -85,7 +85,7 @@ export default function (type, config) {
       return new Promise((resolve, reject) => {
         try {
           storage.setItem(key, string)
-          localSetImmediate(() => {
+          setImmediate(() => {
             cb && cb(null)
             resolve()
           })
@@ -99,7 +99,7 @@ export default function (type, config) {
       return new Promise((resolve, reject) => {
         try {
           storage.removeItem(key)
-          localSetImmediate(() => {
+          setImmediate(() => {
             cb && cb(null)
             resolve()
           })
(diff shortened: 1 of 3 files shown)


rt2zz:
My (limited) understanding is that we need to use `global` to be compatible across all environments (as setImmediate is not always in scope).

If there is an alternative I am all ears.

rt2zz:
Turns out I had it backwards, non-babel users (e.g. ember) have an issue because they have no `global` polyfill. Continuing discussion in #329

Andarist:
how ember tool chain looks like? im completely unfamiliar with it, so cannot provide any good insight 

> What do you think about getting rid of setImmediate from the asyncLocalStorage entirely? It was originally added to make the api actually async, but I am beginning to think this is irrelevant and moreover it just adds a few bytes for no real gain.

Dunno rly, my usage of the lib is pretty basic and only for web, so also - not familiar that much with the codebase and goals to make any opinions on this

toranb:
@Andarist thanks for making the decision to support users w/out global.setimmediate. Just a quick update about the ember tool 

ember-cli does use babel under the hood but without an explicit opt in this global was not working in my experience. I was also trying to keep the config minimal for ember users who adopt this + redux offline (so likely my fault for making that explicit decision).

All that said, I was able to work around the global issue by doing a [rewrite](https://github.com/ember-redux/ember-redux-persist-shim/blob/master/index.js#L20-L21) at build time using broccoli (the file system build tool primitive underlying ember-cli)

It seems that when the next version is published I can throw this out because you "fall back" to `setTimeout` gracefully (given PR #336). Thanks for choosing to support users who are not on the usual React /Webpack tech stack :) I've really enjoyed using persist (and redux offline). Without great community libraries like this I'd be writing a lot more of the foundational code myself and very likely getting it wrong. Thank you !!!

Andarist:
No problem, I've noticed this issue by tweaking my build setup in a work project and ive observed that bunch of libraries cause polyfills for `setImmediate`, `process`, `global` etc to get included in the final build, so I've filed issues and helped with few PRs. Even `react` had this 'issue'.

The whole thing is caused by some implicit auto-polyfilling behaviour of webpack's. At the same time `rollup` and like you have pointed out also the ember tool chain dont cause this.

There is no middleground here at the moment I think, as by guarding against such references we cause polyfills for webpack's users and by not guarding against them we prevent usages outside of webpack.

rt2zz:
ya thats unfortunate. Fortunately we have setTimeout as a reasonable fallback, but this seems like a fundamental compatibility bug in webpack :(

PR #335, opened 2017-04-25
Make redux-persist/constants and redux-persist/storages use the right…

Andarist:
… builds

Basically previous solution forced bundlers to use files from the `lib` directory which has modules transpiled, so it prevented tree-shaking for example.

More can be found [here](https://github.com/redux-saga/redux-saga/pull/895) and in mentioned issues there.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -1,1 +1,0 @@
-module.exports = require('./lib/constants')
--- a/constants/package.json
+++ b/constants/package.json
@@ -1,0 +1,7 @@
+{
+  "name": "redux-persist/constants",
+  "private": true,
+  "main": "../lib/constants.js",
+  "module": "../es/constants.js",
+  "jsnext:main": "../es/constants.js"
+}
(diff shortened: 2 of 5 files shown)


rt2zz:
Thanks, I will need to try this out. Can you clarify my understanding:

`import constants from 'redux-persist/constants'` will resolve to `redux-persist/es/constants`
while
`const constants = requre('redux-persist/constants')` will resolve to `redux-persist/lib/constants`

?

Andarist:
pretty much yes, `webpack2` and `rollup` will go to `es/constants`, while tools like `webpack1` will get contents of `lib/constants`

ive tried it already in `redux-saga` package and it worked great, was trying to figure out how to do this correctly for weeks and suddenly somebody popped up and pointed me into right direction :)

rt2zz:
need to update the `standard` script not to process these files. Or it you want to wait, I think I will be landing prettier in the near future.

Andarist:
Damn! i was preparing this PR before sleep and I forgot to add the most important thing - new files ;) gonna update in a minute

rt2zz:
ha ok this makes more sense, I thought there was some weird module resolution rules that I did not know about 👍

PR #340, opened 2017-04-28
flowtype fixes

sebcode:
Various flowtype fixes, also adds a type for autoRehydrate.


rt2zz:
thank you!

Ashoat:
Awesome! When we can expect to see a new NPM package pushed? I'm getting Flow errors without this patch.

rt2zz:
just released 4.8 👍

PR #341, opened 2017-04-30
Switched to Rollup for UMD builds

### LINKED ISSUE #337: Use rollup for UMD builds
It will produce a smaller bundles, a little bit faster (no runtime code). It also has a nice feature to dedupe babel helpers (such as `extends` which is inlined in each file using object spread). This is no real gain I think as UMD builds usage is not that wide (I think), but its something nice to have. 

Would like to achieve same deduping of helpers for babel when it creates `es` and `lib`, but dont know how to configure it yet.

Andarist:
```
redux-persist.js - 32K -> 25K
redux-persist.min.js - 12K -> 9.1K
```
Needs to be tested. Also I think webpack's lodash polugin generated a little less for `isPlainObject` but dont know any alternative to it for rollup.

Fixes #337


rollup.config.js
```diff
@@ -0,0 +1,50 @@
+import nodeResolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
+import babel from 'rollup-plugin-babel';
+import replace from 'rollup-plugin-replace';
+import uglify from 'rollup-plugin-uglify';
+
+var env = process.env.NODE_ENV
```
marvinhagemeister: Let's use `const` here, to be consistent

rollup.config.js
```diff
@@ -0,0 +1,50 @@
...
+var config = {
```
marvinhagemeister: same as above

package.json
```diff
@@ -14,8 +14,8 @@
     "build": "npm run build:commonjs && npm run build:es && npm run build:umd && npm run build:umd:min && npm run copy-types",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
     "build:es": "cross-env BABEL_ENV=es babel src --out-dir es",
-    "build:umd": "cross-env BABEL_ENV=commonjs webpack",
-    "build:umd:min": "cross-env BABEL_ENV=commonjs NODE_ENV=production webpack",
+    "build:umd": "cross-env BABEL_ENV=es NODE_ENV=development rollup -c -i src/index.js -o dist/redux-persist.js",
```
marvinhagemeister: We could add the entry field in `rollup.config.js` itself. That would make this line a bit tidier:

```js
// rollup.config.js
export default {
  entry: "es/index.js",
  // ..
}
```

marvinhagemeister:
That's a neat size improvement. Thanks for tackling this one 👍  I made a few (really) minor comments, but really it's a huge improvement. Great work!

Andarist:
@marvinhagemeister ive seen u have also put ` "browser": "dist/redux-persist.min.js",` in ur PR, should I do it too? im not familiar with the issue it was fixing so i aint sure

package.json
Andarist: the entry point should be `src/index.js`, i can add this into the config, but personally prefer having those as input/output options in the command line, can be easily copied and built into somewhere else, dont care about this too much though, so @rt2zz's call :P

marvinhagemeister: yeah, I guess it's more a personal preference. You're right that it should point to `src/` 👍  We'll wait for @rt2zz's call on wether we prefer the entry passed via the config or as cli argument.

marvinhagemeister:
@Andarist I think it makes more sense to make a separate PR for that. I was hoping to fix the [unpkg link](https://unpkg.com/redux-persist@4.7.1) which points the the `package.json`'s main field. But that is a different issue. Wasn't right of me to put that in the same PR.

package.json
rt2zz: having not used rollup before, I like it as cli argument - makes it clear what is happening.

rt2zz:
awesome, crazy that we can get a full ~25% savings

marvinhagemeister:
Yeah, rollup really does wonders. A nice bonus is that the output is very human readable compared to webpack 🎉  Thanks again @Andarist for putting in the necessary research and making this PR 💯

Andarist:
Not a problem, was kinda copy pasting most of it as I was converting lately some other projects to the rollup too

PR #342, opened 2017-05-01
Switch to rollup for umd builds. Fixes #337

### LINKED ISSUE #337: Use rollup for UMD builds

marvinhagemeister:
- [x] Make sure `babel-helpers` are only included once
- [x] hopefully fix unpkg package by using the `browser` field in `package.json`


marvinhagemeister:
Whoops, closing in favor of #341 . Didn't saw that one

PR #351, opened 2017-05-12
Deduplicate complete call in getStoredState

Lucretiel:


Andarist:
Could u provide an example when `complete` is called more than once? Your fix might work by accident for sync storages (like `localStorage`), but the library is supposed to work with async ones too - so by the time you call `complete` the keys wont be restored yet.

Lucretiel:
Ah, I didn't realize that getAllKeys could be async.

PR #357, opened 2017-05-21
Added sync storages

Andarist:
Didnt yet tested it properly, so dont merge in. 

Also - those `try/catch`es in `asyncStorage` seems excessive as storage support is already tested by `hasStorage`, dont u think? Or is it fixing some problem?

### DIFF
--- a/src/createPersistor.js
+++ b/src/createPersistor.js
@@ -1,5 +1,5 @@
 import { KEY_PREFIX, REHYDRATE } from './constants'
-import createAsyncLocalStorage from './defaults/asyncLocalStorage'
+import createAsyncLocalStorage from './storages/asyncLocalStorage'
 import purgeStoredState from './purgeStoredState'
 import stringify from 'json-stringify-safe'
 
--- a/src/defaultStorages.js
+++ b/src/defaultStorages.js
@@ -1,4 +1,0 @@
-import createAsyncLocalStorage from './defaults/asyncLocalStorage'
-
-export const asyncLocalStorage = createAsyncLocalStorage('local')
-export const asyncSessionStorage = createAsyncLocalStorage('session')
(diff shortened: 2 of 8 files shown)


rt2zz:
nice, I think we should deprecate `asyncLocalStorage` in v5.

try/catch might still be useful to catch errors such as when quota is exceeded.

Andarist:
good point on the quota exceeded, so do u plan to default to sync web storages + keeping the storage callback API so async ones can get plugged in?

rt2zz:
afaik `setItem` is the only call that errors under normal circumstances, but that would probably warrant more research before committing to dropping try/catch.

Correct we need to keep callback api for the general case, e.g. localforage and AsyncStorage (react-native) are both callback based.

Andarist:
At the moment I supply own `localStorage` wrapper (similar to the one in the PR) with fallback to self-written `memoryStorage` (with exact `localStorage` API). Gonna report back If I get any error reports from sentry regarding this - I've wrapped only `setItem` with `try/catch`.

mrlundis:
What's the status on this? Would love to see sync storage support soon as it would eliminate most of my hard to solve rehydration headache :) 

Also, seems that a comma is missing [on line 14 in syncStorage.js](https://github.com/rt2zz/redux-persist/pull/357/files#diff-417622df8208b84eaac84e6cf4e66c27R14).

Andarist:
thanks, fixed :)

rt2zz:
v5 incorporated this into it, default storages are sync.

The rehydration process however is still async, and now there is a `PersistGate` component that helps manage delaying of rendering until rehydration is complete. Thanks all 👍

Andarist:
@rt2zz I've recently published [this little lib](https://github.com/Andarist/memory-web-storage/blob/master/src/index.js) if you'd like to reuse some logic, i.e. storage availability test

PR #359, opened 2017-05-26
Added prepare script, which can be used by npm5 when the package is u…

Andarist:
…sed as git dependency

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,8 @@
     "build:umd": "cross-env BABEL_ENV=es NODE_ENV=development rollup -c -i src/index.js -o dist/redux-persist.js",
     "build:umd:min": "cross-env BABEL_ENV=es NODE_ENV=production rollup -c -i src/index.js -o dist/redux-persist.min.js",
     "build:watch": "npm run build:commonjs -- -watch",
-    "prepublish": "npm run clean && npm run build"
+    "prepare": "npm run build",
+    "prepublish": "npm run clean && npm run prepare"
   },
   "ava": {
     "require": [


package.json
```diff
@@ -17,7 +17,8 @@
     "build:umd": "cross-env BABEL_ENV=es NODE_ENV=development rollup -c -i src/index.js -o dist/redux-persist.js",
     "build:umd:min": "cross-env BABEL_ENV=es NODE_ENV=production rollup -c -i src/index.js -o dist/redux-persist.min.js",
     "build:watch": "npm run build:commonjs -- -watch",
-    "prepublish": "npm run clean && npm run build"
+    "prepare": "npm run build",
```
marvinhagemeister: Do we actually need the `prepare` script? The [docs](https://docs.npmjs.com/misc/scripts) says that this will run both before publishing and `npm install`. The latter seems to be not necessary because the package already includes the compiled `lib/` folder.

Andarist: According to http://blog.npmjs.org/post/161081169345/v500
> Git dependencies with prepare scripts will now have their devDependencies installed, and their prepare script executed as if under npm pack.

Its important as entry points to the lib are [.gitignored](https://github.com/rt2zz/redux-persist/blob/2d3f1e4cdb930b275a5ed2d10faebf26be6358ab/.gitignore#L2-L4), so if somebody forks and want to use its fork from git its rly inconvenient now (before npm5 running `prepare` in such case), because he needs to:
- probably alter entry points (`main`)
- probably unignore `es`/`lib` 
- probably run build by himself and push generated files to the git

marvinhagemeister: Oh I see, I didn't know about that use case. Makes sense 👍

rt2zz:
nice! npm finally adding proper git install support!!

thanks @Andarist

Issue #303, opened 2017-03-21
Is it possible to create an async transform?


cascornelissen:
Looking through the listed transforms I've noticed [redux-persist-transform-encrypt](https://github.com/maxdeviant/redux-persist-transform-encrypt) has an async version but it appears to simply wrap the sync one and the README has the following note:

> Asynchronous support is still a work in progress.

Any plans on supporting async transforms in the (near) future?

cascornelissen:
@rt2zz, I would love an answer on this as I will have to write my own storage solution or submit a PR if this isn't supported in redux-persist. Thanks in advance!

rt2zz:
@cascornelissen sorry for the delay. Currently async transforms are not supported. I am open to the idea but we would need to be very careful around performance and also error handling.

What is the specific need that is async? Is it possible you can bake the async portion into the storage engine (which is already async, namely `getItem` and `setItem`)

cascornelissen:
@rt2zz, I'm trying to transform data via [SubtleCrypto](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto) before storing it.

I started working on something but haven't had the time to dive into it any deeper. You might want to give me some pointers/tips though, I've only implemented a working inbound transform so far: https://github.com/rt2zz/redux-persist/compare/master...cascornelissen:master

rt2zz:
nice, approach seems reasonable. I am still not sure how this feature fits in, given that this complicates the common case. I am working on some lower level refactor atm, so for the time being I think the best options are either
- keep rolling with your fork
- implement the async encrpytion in the storage engine

cascornelissen:
Give the fact that I want to use multiple storage engines based on the environment later on I'll go with option 1 for now. Could you let me know when this refactor is finished (and do you have any ETA)? I might want to submit a PR at that time to implement the Promise-based transforms if anyone's interested.

For the argument saying that it complicates the common case: I agree, maybe we could make this configurable behind a flag? But we'll see once your refactor is finished.

CharlieHess:
@cascornelissen I actually need the same thing; did you keep rolling with your fork (and is it published anywhere?)

CharlieHess:
@rt2zz I went ahead and implemented it in https://github.com/rt2zz/redux-persist/pull/360. I have a pretty well-defined use case (a reducer that is used very infrequently, but needs to imbue state with passwords from https://github.com/atom/node-keytar) and in my testing there is no perf impact.

[...]

PR #360, opened 2017-05-28
Async transforms

CharlieHess:
This PR resolves https://github.com/rt2zz/redux-persist/issues/303 by adding support for async transforms, behind a configuration flag (default behavior is unchanged). Folks with a well-defined use case can opt into this behavior at their own risk.


CharlieHess:
In [my case](https://github.com/CharlieHess/redux-persist-transform-passwords) I wanted to store secrets securely (say, in the keychain) and have them excluded from the persisted shape. Then, on rehydrate, I wanted to read the secrets from the keychain and apply them to the state as if they were there all along.

CharlieHess:
@rt2zz any thoughts on this one?

antondomratchev:
@rt2zz We have the exact same use case for this feature, this would be very nice to have in the package.

rt2zz:
@CharlieHess sorry for the delay on getting back here. I like the implementation but am concerned by the raw amount of code, and also the possible support burden. I have been working on redux-persist v5 lately and there might be an opportunity to make some more drastic changes there.

Some options
1. support async transforms out of the box, change the time iterator to no longer use setInterval and instead wait for the result of each transform before continuing to the next piece of state (potentially with a setTimeout in between)
2. in v5 while we serialize in chunks, we aggregate the results and only write to disk once. With this, instead of trying to do async at the transform level, we could consider a new hook to pre/post process the aggregated state just before writing to disk and just after reading from disk.

Of course the challenge with option 2 is the variations are endless: some use cases might prefer to process the state *before* serialization, and other *after*. Some might require that the async process is blocking while others can run in parallel.

Since this current solution is working for you I encourage you to keep using it for the time being (redux-persist@4 is not going to change much at all, so keeping up to date should not be difficult) while we take more time to think through how the api should work in v5.

mnquintana:
@rt2zz 👋  Just checking in - now that v5's out, what do you think would be the best approach for supporting async transforms now?

sebastian-schlecht:
any progress on this?

pitops:
Guys I really need this, any progress on merging?

CharlieHess:
We haven't needed to update to v5 yet and thus our fork (based off this branch) is still serving us for the time being. I won't be able to get back to this anytime soon but I bet Zack would consider a pull request!

pirtlj:
@rt2zz any progress here?  I would love to see this supported.

PR #372, opened 2017-06-14
Update type definition for getStoredState

cbayram:
`getStoredState` return type changed to `Promise<any>`

 This helps avoid typescript error "Operand for 'await' does not have a valid callable 'then' member.." when using with await; e.g. `await getStoredState(persistConfig)`


rt2zz:
I believe it should be `Promise<State>`?

cbayram:
I don't think `State` is a defined type in this case.

rt2zz:
ah ya I was mistaken, at least in flow we would have no good way to parameterize that. Perhaps `Promise<any>` was in fact correct? I am not a TS user :/

marvinhagemeister:
@cbayram Thanks for your PR 🎉  Using generics (the `State` in) `Promise<State>` is the right way to do it.

Issue #379, opened 2017-06-21
Not working with webpack3


lizux:
Console Error log: TypeError: 'setTimeout' called on an object that does not implement interface Window.

rt2zz:
more discussion https://github.com/rt2zz/redux-persist/issues/366

open to PR's, I am not currently using webpack 3 and cannot weigh in on the specifics

PR #380, opened 2017-06-21
Fixed illegal invocation, caused by setTimeout being assigned as obje…

### LINKED ISSUE #366: Stopped Working
Not really sure why or what happened - woke up this morning and updated webpack and redux-persist now no longer works.  Says illegal invocation when trying to call setImmediate()

```
persistStore.js:25 Uncaught TypeError: Illegal invocation
    at persistStore (persistStore.js:25)
```

image

Andarist:
…ct property in webpack3 (fixes #366 and #379)

@rt2zz this should fix issues with webpack3, needs to be published.


rt2zz:
❤️ release going out now

PR #385, opened 2017-06-29
Note that custom storage must support callbacks

XertroV:
Spent about an hour trying to figure out why redux-persist wasn't working, and after plenty of debugging I realised it was because custom storage modules need to use callbacks, not promises. Added this factoid to the readme.


rt2zz:
ty!

Issue #386, opened 2017-06-29
Break IE from 4.7+


jalkoby:
Hi. After update from 4.6 to 4.8.1 my app had broken in all IE browsers(10, 11, edge) with "". I've made a short investigation and found the place - src/utils/setImmediate.js . In that file the code tries to use a native `setImmediate` which only present in node & IE. Probably the best case is avoid native usage for ie. What do you think and should I create a PR?

Andarist:
What's the exact issue with `setImmediate` in IE?

jalkoby:
It raises "Invalid calling object"  in: 
```js
# src/persistStore.js

if (shouldRestore) {
    setImmediate(() => {
```

One of possible reasons is https://github.com/facebook/draft-js/issues/1230

Andarist:
Could u change `src/utils/setImmediate.js` (in reality u need to change appropriate file which is in `lib/` or `es/` directory - depending on which bundler do you use) to:
```js
const setImmediate = typeof global !== 'undefined' && typeof global.setImmediate !== 'undefined'
  ? (fn, ms) => global.setImmediate(fn, ms)
  : (fn, ms) => setTimeout(fn, ms)
```

I believe it should fix your issue, would like to get a confirmation, as I cannot check it at the moment myself. If you confirm it's fixed - we are going to incorporate it into the library.

jalkoby:
@Andarist yep, wrapping helped to fix it - here is the PR

PR #387, opened 2017-06-29
Fix IE support

jalkoby:
The fix for #386


src/utils/setImmediate.js
```diff
@@ -1,3 +1,4 @@
-const setImmediate = typeof global !== 'undefined' && typeof global.setImmediate !== 'undefined' ? global.setImmediate : (fn, ms) => setTimeout(fn, ms)
+const hasNativeSupport = typeof global !== 'undefined' && typeof global.setImmediate !== 'undefined'
+const setImmediate = hasNativeSupport ? (fn, ms) => global.setImmediate : (fn, ms) => setTimeout(fn, ms)
```
Andarist: returned functions returns `global.setImmediate` instead of calling it with provided arguments ;)

jalkoby:
Sorry - rush fixing (without running test suite at the end) is not a good idea

Andarist:
Thanks! Cannot publish it though yet, needs to be done by @rt2zz

Andarist:
@jalkoby just released with a fix, please check it out :)

jalkoby:
thanks

PR #390, opened 2017-07-04
Call resume even if rehydrateAction throws an exception.

pavoni:
Tracked down a bug in my code where an exception was being thrown as a consequence of a persisted state being loaded, this stopped redux persist working.

Turned out that the persist store was being left in a paused state.

Quite a lot can happen as a result of `store.dispatch(rehydrateAction(restoredState, err))`, so I think we should make sure that the `persistStore` is resumed if there is an exception.

I thought it was also useful to trace the error - but this is perhaps less of an issue.

### DIFF
--- a/src/persistStore.js
+++ b/src/persistStore.js
@@ -30,8 +30,6 @@ export default function persistStore (store, config = {}, onComplete) {
         }
         try {
           store.dispatch(rehydrateAction(restoredState, err))
-        } catch (err) {
-          console.log(err)
         } finally {
           complete(err, restoredState)
         }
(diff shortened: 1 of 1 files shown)


src/persistStore.js
```diff
@@ -28,9 +28,13 @@ export default function persistStore (store, config = {}, onComplete) {
           if (purgeKeys === '*') restoredState = {}
           else purgeKeys.forEach((key) => delete restoredState[key])
         }
-
-        store.dispatch(rehydrateAction(restoredState, err))
-        complete(err, restoredState)
+        try {
+          store.dispatch(rehydrateAction(restoredState, err))
+        } catch (err) {
+          console.log(err)
```
rt2zz: lets remove the console.log

pavoni:
@rt2zz If this looks OK are you happy to merge?

pavoni:
@rt2zz Thanks for merging. 

Can see you're busy working on v5 - but would be love another release of 4 when you get a moment - so I can drop my branch.

rt2zz:
👍 @4.9.0 just released

pavoni:
Perfect - thanks very much.

PR #406, opened 2017-08-07
(persistStore): Add promise support

isnifer:
Hi! I try to find a solution for Delayed render while `redux-persist` restore app state. Got this - https://github.com/rt2zz/redux-persist/blob/master/docs/recipes.md.

For now, my `store.js` is so big, I want to keep store initialization in the separate file.
I thought export `promise` property from `persistStore` would be a great feature like a `onComplete` callback.

#### store.js
```js
// ...

export const persistor = persistStore(store, { ... })

export default store
```

#### App.js
```js
// ...
import store, { persistor } from './store'

export default class App extends PureComponent {
  state = {
    loaded: false,
  }

  componentWillMount() {
    persistor.promise.then(() => {
      this.setState({ loaded: true })
    })
  }

  render() {
    if (!this.state.loaded) {
      return (
        <View style={{ backgroundColor: '#000000', flex: 1 }}>
          <Text style={{ color: 'pink', fontSize: 45 }}>Loading</Text>
        </View>
      )
    }

    return (
      <Provider store={store}>
        <View style={styles.container}>
          <Navigator />
          <InitialScreen />
        </View>
      </Provider>
    )
  }
}
```

### DIFF
--- a/src/persistStore.js
+++ b/src/persistStore.js
@@ -15,39 +15,43 @@ export default function persistStore (store, config = {}, onComplete) {
   const persistor = createPersistor(store, config)
   persistor.pause()
 
-  // restore
-  if (shouldRestore) {
-    setImmediate(() => {
-      getStoredState(config, (err, restoredState) => {
-        if (err) {
-          complete(err)
-          return
-        }
-        // do not persist state for purgeKeys
-        if (purgeKeys) {
-          if (purgeKeys === '*') restoredState = {}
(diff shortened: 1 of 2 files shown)


rt2zz:
this is now handled via `PersistGate` in v5. The idea behind that is a component that can go around your app (right along side Provider):
```js
  <Provider store={store}>
    <PersistGate persistor={persistor} loading={<div>Loading...</div>}>
```

PR #413, opened 2017-08-15
Enable multiple persistReducer calls to work correctly

### LINKED ISSUE #412: [v5] Multiple persistReducer causes failure during REHYDRATE
To illustrate, if I have a root reducer set up as below, the user store will be rehydrated correctly, but the incidents will not. Two `REHYDRATE` actions are dispatched, and the first (user) results in a state change, while the second results in no state change. 

```js
const rootReducer = combineReducers({
	...otherReducers,
	location: reducer,
	user: persistReducer({ key: 'user', storage }, user),
	incidents: persistReducer({ key: 'incidents', storage }, incidents)
})
```

itslittlejohn:
I believe this fixes #412. I was debugging this and observed that on line 99 you state: 

`@NOTE if key does not match, will continue to default case`

This is inside `case REHYDRATE`.  For some reason, this case was falling through after the first REHYDRATE call in some instances, but it doesn't fall through to `default`. Instead it falls through to `PURGE`, which I don't think was the intent. 

If you take the example I gave in the related issue, `combineReducers` maps each action to each reducer. So first `action.key (user) === config.key (user)` but in the second pass, `action.key (user) === config.key (incidents)` and it falls through and set's `_purge` to true. 

Then, when it goes to process the second `REHYDRATE` (`action.key (incidents)`), we never get to line 100 because of `if (_purge) return state`.

So it seems that _purge isn't being reset across `REHYDRATE` actions as well?

So all that to say, I know that this change fixes my use case of wanting to REHYDRATE multiple store nodes, but I'm not sure if this change is the actual final solution or just a first step.

### DIFF
--- a/src/persistReducer.js
+++ b/src/persistReducer.js
@@ -14,7 +14,7 @@ import getStoredState from './getStoredState'
 import purgeStoredState from './purgeStoredState'
 
 type PersistPartial = { _persist: PersistState }
-/* 
+/*
   @TODO add validation / handling for:
   - persisting a reducer which has nested _persist
   - handling actions that fire before reydrate is called
@@ -113,11 +113,6 @@ export default function persistReducer<State: Object, Action: Object>(
           }
         }
 
-      case PURGE:
-        _purge = true
-        purgeStoredState(config)
-        return state
-
       default:
         // @TODO more performant workaround for combineReducers warning
         let newState = {
@@ -126,6 +121,11 @@ export default function persistReducer<State: Object, Action: Object>(
         }
         _persistoid && _persistoid.update(newState)
         return newState
+
+      case PURGE:
+        _purge = true
+        purgeStoredState(config)
+        return state
     }
   }
 }
--- a/src/persistReducer.js
+++ b/src/persistReducer.js
@@ -113,19 +113,19 @@ export default function persistReducer<State: Object, Action: Object>(
           }
         }
 
-      default:
-        // @TODO more performant workaround for combineReducers warning
-        let newState = {
-          ...baseReducer(restState, action),
-          _persist,
-        }
-        _persistoid && _persistoid.update(newState)
-        return newState
-
       case PURGE:
         _purge = true
         purgeStoredState(config)
         return state
     }
+
+    // @NOTE default pass through reducer if the switch statement above did not return
+    // @TODO more performant workaround for combineReducers warning
+    let newState = {
+      ...baseReducer(restState, action),
+      _persist,
+    }
+    _persistoid && _persistoid.update(newState)
+    return newState
   }
 }


rt2zz:
thanks, that was definitely an error. Putting purge below default will simply cause purge never to be called, I think rather what we need is to explicitly call the default action when the key does not match. Simplest code would probably be to switch to if statements instead of a switch.

itslittlejohn:
Okay, I think I'm following. But if we just returned the default action (something like below maybe?) wouldn't we still never be falling through to `PURGE`? I guess I assumed that the `PURGE` action was fired explicitly, and if that was the case, I would have expected it to catch even if `default` comes before in the switch statement. Correct me if I'm wrong.

```js
case REHYDRATE:
  // noop if purging
  if (_purge) return state

  // @NOTE if key does not match, will continue to default case
  if (action.key === config.key) {
    let reducedState = baseReducer(restState, action)
    let inboundState = action.payload
    let reconciledRest: State = stateReconciler(
      state,
      inboundState,
      reducedState,
      config
    )

    return {
      ...reconciledRest,
      _persist: { ..._persist, rehydrated: true }
    }
  } else {
    return defaultAction()
  }

case PURGE:
  _purge = true
  purgeStoredState(config)
  return state

default:
  return defaultAction()
```

rt2zz:
ya, actually how about this for a clean solution: lets just move default out of the switch entirely:
https://gist.github.com/rt2zz/2eddff64ddf258412760a5328bc0c99a

itslittlejohn:
That's way better. :)

rt2zz:
thanks for catching this, big oversight on my part! Do you mind updating the PR accordingly?

itslittlejohn:
Not at all!

itslittlejohn:
Sorry.. was just about to get to it now. Looks like we're all set!

rt2zz:
all good 👍 just published alpha.3

itslittlejohn:
You know what.. I've just got around to testing this, and it's still failing in the same way. I may not fully understand what you're doing here with `PURGE`. It's still falling through to PURGE for the second reducer because `action.key === config.key` is false. It would have matched the second time through but purge was set so we do the noop. 

Does that make sense? I guess I didn't look at the change closely enough. Sorry about the merry-go-round on this. :)

rt2zz:
my bad, needed a break in there :(

I rewrote it with if/else as I think that will prevent confusion in the future at the cost of a tiny bit of readability. https://github.com/rt2zz/redux-persist/blob/v5/src/persistReducer.js#L54

Lastly added tests and published as alpha 5. Thanks for help working through this.

itslittlejohn:
That should do it! I'll try it out tonight for redundancy.

PR #416, opened 2017-08-19
fix error where persistReducer function doesn't return any value

itslittlejohn:
Just got around to testing, and I found that if the keys don't match, persistReducer doesn't return anything and redux throws an error. 

https://github.com/rt2zz/redux-persist/blob/a4c3695395852ecf20ff186f74c2c598f87fe291/src/persistReducer.js#L95-L124

Removing the `else` and letting it truly fall through fixes it for me. 

Third time's the charm? ;) On the plus side, I'm getting much better at github.. ha.

### DIFF
--- a/src/persistReducer.js
+++ b/src/persistReducer.js
@@ -112,15 +112,15 @@ export default function persistReducer<State: Object, Action: Object>(
           _persist: { ..._persist, rehydrated: true },
         }
       }
-    } else {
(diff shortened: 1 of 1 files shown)


rt2zz:
wow ya I do not know what I was thinking. Thanks for you help and patience!

rt2zz:
just published redux-persist@5.0.0-alpha.7

itslittlejohn:
Great! Glad I could help.

PR #421, opened 2017-08-21
[types] mark config as optional in persistStore

cwhenderson20:
According to both the API documentation for `persistStore` and the method definition in the code itself allow for the `config` parameter to be omitted.

This PR modifies the flow type of `persistStore` to align with that expectation.


rt2zz:
thank you!

cwhenderson20:
My pleasure. Thank you for the module! It's invaluable.

rt2zz:
just released v4.9.1 with this. Note: v5 is around the corner, but we do not have typescript definitions for v5 yet. 

Should not be too hard to figure out though as the new code base is now written in flow which should be mostly compatible type wise (nothing fancy)

PR #427, opened 2017-08-27
Support async transforms

Noitidart:
I don't know if we want to support this, but i think the changes I made for this are correct.

### DIFF
--- a/src/createPersistoid.js
+++ b/src/createPersistoid.js
@@ -40,9 +40,12 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
         }
 
         let key = keysToProcess.shift()
-        let endState = transforms.reduce((subState, transformer) => {
-          return transformer.in(subState, key)
-        }, lastState[key])
+        let endState = lastState[key];
+        (async function() {
+          for (const transformer of transforms) {
+            endState = await transformer.in(endState, key);
+          }
+        })();
         if (typeof endState !== 'undefined') stagedWrite(key, endState)
       }, throttle)
     }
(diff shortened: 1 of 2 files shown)


rt2zz:
I am not against this, but I am concerned by a few things: 
1. is async await support prevalent enough to include at this point?
2. what happens if the async transforms return out of order. You might stagedWrite an old state over a newer one. We probably need some sort of nonce to protect against this
3. is there a perf cost to the IIFE? any reason not to simply make the parent callback async?
4. do we need a timeout based write state trigger. Right now staged write waits for all keys to drain before actually writing out state, but async transforms create a risk that the pending key pool infrequently gets fully drained. This is the most performance critical and difficult to debug piece of redux-persist so I want to make sure we get this right.

Noitidart:
Thanks, I'm not sure about that IIFE perf cost. Doesn't this `async` `await` get transpiled out?

Yeah I totally am newbie at redux-persist, my implementation of async transforms is probably wrong, but I thought I covered it so that they don't return out of order.

Noitidart:
As I have been using this over the last couple weeks, I think async transforms are super useful. But I couldn't do it right and am not sure how to fix it up. Pretending if async transforms was done from scratch by someone who knows what they are doing, do you think async transforms will make it into v5 one day?

rt2zz:
:) I still want to figure this out, it does seem to have a lot of value, but also a lot of risk. Once the first release of v5 is out and stable I think will be easier to re-evaluate. As for this PR, I say we leave it open as a reminder to follow up.

Noitidart:
Thanks for that awesomely positive and hopeful comment!!

aguynamedben:
Tracking in this issue: https://github.com/rt2zz/redux-persist/issues/303
And this PR: https://github.com/rt2zz/redux-persist/pull/360

PR #439, opened 2017-09-02
Added unpkg field in package.json

Andarist:
This will allow nicer unpkg urls.

https://unpkg.com/redux-persist instead of https://unpkg.com/redux-persist/dist/redux-persist.min.js


rt2zz:
great, I still need to get around to applying a rollup umd build to v5

Issue #252, opened 2017-01-15
Is there some way to ensure data is flushed into storage?


littlesome:
I' m saving user's login state using this library, when user press the logout button, my app do this:

1. dispatch an action to clear the login state in redux
2. wait for 200ms, then reload the app

wait for 200ms is not good, if there's a callback when pending operations are finished would be better

rt2zz:
There is not presently a way to do this. a flush option has been discussed.

Is there a reason you want to reload the app as opposed to sending out some kind of `RESET` action that resets state to initial state?

Phredward:
Hi.  I would like to restart my react-native app to apply a code-push update.  Right before I do this, I store state in redux, and it is not persisted (unless I wait, which is an acceptable workaround for now).

I would be happy to check a variable in redux, call a method that tells me if my store is persisted, or have an alternative restart method that redux provides that finishes syncing and then restarts.

This raises another concern, what prevents a user from killing the app quickly after state is updated (besides the limits of human finger speed)?

goodhyun:
I'd also like to make sure that every bits are persisted when the app is shutting down.
How's the progress of #217 by @simonacca ?

rt2zz:
in practice I have not observed unsaved state to be an issue. That said I do like the idea of having a `flush` method which write all keys simultaneously and returns a promise. As long as this could be implemented cleanly and without adding a many bytes of code.

Klathmon:
It's not a long-term solution, but I've got a "hack" that can let you at the very least get confirmation that your store was persisted before doing anything that requires it (like a page refresh).

In my case I'm using `localforage` as my store, so I just "proxy" the `setItem` key on the localforage object to be my own function which I can then use to run whatever I want once the `setItem` in localforage is complete. Here's the snippet that hopefully explains what I'm doing better than my gibberish:

```
const localforageSetItem = localforage.setItem.bind(localforage)

localforage.setItem = (key, value, callback) => {
  return localforageSetItem(key, value, (...args) => {
    console.log('Done persisting!')
    callback(...args)
  })
}
```


I'm going to use it to fire custom events so I can attach a listener to get notified when a store is fully persisted and it's safe to move on. It's not a particularly good solution, but it will work until either a manual persist gets implemented, or someone comes up with something much less horrible.

rt2zz:
Thinking about this more, I would be favor of a `flush` method on the persistor which immediately writes all pending storage operations and returns a promise.

rt2zz:
[...]

PR #445, opened 2017-09-08
Flush function

Klathmon:
For issue #252 

This adds a `flush` function to the persistor object which when called will force everything currently in the `storesToProcess` to get flushed to the storage engine, and it returns a promise that resolves when completed.

It looks like a bunch of things are different in the "changed files" but most of that is because I had to pull out some inline files into their own defined functions.

I had to pull the function in the `subscribe` call out and call it during the flush because of how Redux handles multiple nested calls to `dispatch` (it doesn't call the subscribed listener until the "root" dispatch call completes).

So if you had multiple nested calls to `dispatch` in your application and one of the deeper ones calls this `flush` function, you may not have had some of the stuff dispatched to the store flushed to the storage engine. Calling `storeListener` in the flush function solves this problem.

Let me know if you want me to change anything or explain my reasoning. I also tried to keep it in your style, and it passes standard.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,6 +77,8 @@ persistStore(store, config, callback).purge()
       - pauses persistence
     - `resume()`
       - resumes persistence
+    - `flush()`
+      - forces everything to be written to storage. Returns a promise that resolves when everything is stored successfully, rejects on any errors.
 
 #### `autoRehydrate(config)`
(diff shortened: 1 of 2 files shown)


Klathmon:
Ugh... I just noticed #321 which has tests and better documentation, so I'm just going to close this PR.

PR #446, opened 2017-09-11
Fix typescript typings for Storage

cliffkoh:
This fix is per the discussion at https://github.com/Microsoft/TypeScript/issues/18397. I'm positive this typing is the actual intent of the generic `Result` :)

### DIFF
--- a/type-definitions/index.d.ts
+++ b/type-definitions/index.d.ts
@@ -1,24 +1,10 @@
 declare module "redux-persist" {
   import { Store, StoreEnhancer } from "redux";
 
-  export interface AsyncStorage<Result> {
-    setItem(key: string, value: Result, onComplete?: OnComplete<Result>): Promise<void>;
-    getItem(key: string, onComplete?: OnComplete<Result>): Promise<Result>
-    removeItem(key: string, onComplete?: OnComplete<Result>): Promise<void>;
-    getAllKeys(onComplete?: OnComplete<Result[]>): Promise<Result[]>;
-  }
-
-  export interface WebStorage<Result> {
-    setItem(key: string, value: Result, onComplete?: OnComplete<Result>): void;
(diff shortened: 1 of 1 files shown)


type-definitions/index.d.ts
```diff
@@ -1,19 +1,18 @@
...
     ): Promise<Result | string>;
     removeItem(key: string, onComplete?: OnComplete<any>): Promise<any>;
-    getAllKeys?<Result>(
+    getAllKeys?(
       callback?: (error?: Error, keys?: string[]) => void
     ): Promise<string[]>;
-    getAllKeys?<Result>(onComplete?: OnComplete<Result>): Promise<Result>;
+    getAllKeys?(onComplete?: OnComplete<Result>): Promise<Result>;
     keys?: (...args: any[]) => any;
-    [key: string]: any; // In case Storage object has some other (private?) methods and properties.
```
cliffkoh: This is not needed for an interface - the concrete Storage object would be of the right type (e.g. AsyncStorage) and able to access those other methods and properties, while redux-persist does not care for the existence of those other methods and properties for it only relies on the methods/properties defined here in the interface.

cliffkoh:
I took a closer look at the type definitions and noticed that there was a bit of a hack added to make it work with AsyncStorage (I was using an older version where I also hit the same problem as @iRoachie). I cleaned that up and also make the signatures compatible with some of the variants of StorageAPIs.

@iRoachie it would be great if you can review the change. I tested it and it works fine with AsyncStorage for me (pass in String to the generic).

iRoachie:
Good job mate, I was meaning to do it at the time. Just by looking at it, it looks good

rt2zz:
I am not a TS user, but this looks good to me. If anyone finds issue please open.

cliffkoh:
@rt2zz any chance you can release a new patch build to npm which picks up this change? :)

PR #448, opened 2017-09-13
Warn if stringify fails in development

scotchapman:
In development (non-production) environment,  console.warn the error message.


rt2zz:
woah, this looks like a bug. I am leaning towards lets be consistent across all env, I am just not sure is it better to throw or return null?

rt2zz:
closing as no longer relevant in v5, thanks!

PR #453, opened 2017-09-19
avoid race conditions icw persistor.pause

Morriz:
To avoid race conditions after` persistor.pause()` is called.

Atomicity of the entire operation is still preserved.


Morriz:
sorry, somehow the github editor was clumsy and left unwanted spaces :|

PR #455, opened 2017-09-20
Notify when state stored

Froelund:
Add ability to supply a callback to `storage.setItem`. If not supplied, it will use the default.


Froelund:
Travis build is failing, but I believe that to be caused by a flaky test.

rt2zz:
can you elaborate on the need for this? what action do you plan on taking upon storage write? I am wondering if this should be generalized for all storage operations, or if perhaps this should be a feature of the storage engine itself.

PR #466, opened 2017-09-26
fix compatibility with flowtype 0.55.0

sebcode:
Latest flowtype 0.55.0 bails out with "Unexpected token in"


rt2zz:
ty, I updated the labels slightly 👍

PR #481, opened 2017-10-05
Adds redux-persist-react-native-fs to ecosystem.md

jsdario:
Storage engine over react-native-fs by netbeast team https://www.npmjs.com/package/redux-persist-react-native-fs
Tested in production in https://getyeti.co


rt2zz:
thank you, looks like this is likely a dupe of https://github.com/leethree/redux-persist-fs-storage

Nothing wrong with that, although for the sake of the docs I think we will need to settle on just one to include in the readme.

PR #483, opened 2017-10-09
Rehydrate including reduced state

martin-rueegg:
In v4, when auto-rehydrating, the reduced state is enhanced by the incoming state. However, this is no longer the case in v5, if for the given subtree a transform function is provided.

This PR adds a flag that allows previous behaviour.

Happy to add to the docs as well, if PR in general is accepted. In fact, it's more like a RFC.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "redux-persist",
-  "version": "5.0.0-rc.6",
+  "version": "5.0.0-rc.6.bp.3",
   "description": "persist and rehydrate redux stores",
   "main": "lib/index.js",
   "module": "es/index.js",
--- a/src/constants.js
+++ b/src/constants.js
@@ -8,3 +8,5 @@ export const PERSIST = 'persist/PERSIST'
 export const PURGE = 'persist/PURGE'
 export const REGISTER = 'persist/REGISTER'
 export const DEFAULT_VERSION = -1
+export const AUTO_REHYDRATE = 1
+export const AUTO_REHYDRATE_WITH_REDUCED_STATE = 3
--- a/src/persistReducer.js
+++ b/src/persistReducer.js
@@ -132,7 +132,7 @@ export default function persistReducer<State: Object, Action: Object>(
         let reducedState = baseReducer(restState, action)
         let inboundState = action.payload
         let reconciledRest: State =
-          autoRehydrate === false
+            ! autoRehydrate
             ? reducedState
             : stateReconciler(state, inboundState, reducedState, config)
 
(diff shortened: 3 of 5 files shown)


rt2zz:
I am finding there are many possible desired state reconciliation strategies, including this one should state inherit from pre-rehydrate state? But actually the problem is further complicated because this actually creates two levels of "shallow merge", i.e. in the standard combineReducers case where persistReducer is applied to the root reducer:
1. reconcile each sub-reducer that we have incoming state for
2. shallow merge the state for each of these sub reducers

Where this starts to get really tricky is when persistReducer is not applied at the root, but lets say e.g. on just the "auth" reducer. if auth has the shape: `{ user: {...} }` We will end up shallow merging the user object which I think is not desired.

I need to think about this more. Not sure how to fully capture this useful logic without restricting the number of use cases for which reconciliation is useful.

mauron85:
Ok. I understand the problem. But what is the best way how to merge incoming state with initial state.
This is actually problem when using filters.

Let say state atom looks like this:

```
STATE = {
  entities: {
    reducer1: {},
    reducer2: {},
  }
}
```

Then we filter reducer1 to be only persisted.

const entitiesSubsetFilter = createFilter(
  'entities',
  ['reducer1']
);

Current behaviour:
After hydration is completed entities.reducer2 will be undefined, which cause problems in selectors.

Expected behaviour:
After hydration entities.reducer2 (which is initial state) is merged with rehydrated entities.reducer1.

Is is possible to merge those entites? @martin-rueegg mentioned transform function. Can it be used for merging?

mauron85:
Not 100% but I found some interesting behavior.
I've tried to restore state to initial after REHYDRATE action in filtered reducer (see previous comment).

```
export default function roomsReducer(state = {}, action) {
  if (action.type === REHYDRATE) {
    return {};
  }
  return state;
}
```

The interesting fact (for me) is that I have to return new empty state object. Only then entities.rooms will be added to state.

mauron85:
I managed to reproduce this scenario:

https://jsfiddle.net/mauron85/v6tf9xma/

How to reproduce:
1. Click on "Add" button to store some data in localStorage
2. Refresh Browser (or hit Run)
3. Open developer console. There will be error message:

```
Uncaught (in promise) TypeError: Cannot read property 'trackingNr:undefined' of undefined
    at Function.mapStateToProps2 [as mapToProps] (<anonymous>:116:30)
    at r (react-redux.min.js:1)
    at p (react-redux.min.js:1)
    at a (react-redux.min.js:1)
    at react-redux.min.js:1
    at Object.run (react-redux.min.js:1)
    at p.onStateChange (react-redux.min.js:1)
    at Object.a [as dispatch] (redux.min.js:1)
    at Object.rehydrate (redux-persist.js:997)
    at redux-persist.js:338
```

PR #496, opened 2017-10-20
[autoRehydrate] configurable autoRehydrate with default

rt2zz:
This PR updates the autoRehydrate config to take a function, which when provided will be used to reconcile state during rehydration. This will allow us to support a variety of reconciliation strategies, including notably hard set and deep merge.

I am still uncertain on the naming, should the config option be `autoRehydrate` or `stateReconciler`? Also for the default strategy, right now I named it `autoMergeLevel1` which as two properties
1. merges 1 level deep
2. will skip rehydration if substate was modified

Most people will never need to worry about this since it is the default, but I suspect there is a better name. Open to suggestions.


rt2zz:
decided to go with stateReconciler over autoRehydrate

PR #507, opened 2017-10-26
Add option to opt-out of serializing and deserializing state from storage

reblws:
When working with browser extensions the automatic serializing and deserializing with JSON.stringify and JSON.parse before reading & writing to the store can be troublesome.

The [browser.storage.local](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/storage/local) API for extensions already handles parsing state and returns a native object for `.get` calls. In order to get redux-persist to work with storage.local, in my `storage.setItem` function I had to write code calling JSON.parse to get a native object again before writing to storage. This seems pretty unnecessary to me.

### DIFF
--- a/src/createPersistoid.js
+++ b/src/createPersistoid.js
@@ -13,8 +13,8 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
   const storageKey = `${config.keyPrefix !== undefined
     ? config.keyPrefix
     : KEY_PREFIX}${config.key}`
-
   const storage = config.storage
+  const serialize = config.serialize === false ? x => x : defaultSerialize
 
   // initialize stateful values
   let lastState = {}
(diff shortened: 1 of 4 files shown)


rt2zz:
@reblws I apologize for not reviewing this sooner. Is there a reason you closed it other than age? this seems like the right approach to me and fwiw I prefer this implementation to https://github.com/rt2zz/redux-persist/pull/572 for aesthetic reasons

reblws:
Ah, just thought it would make it easier for you if you had to review less PRs.

rt2zz:
thanks for looking out :D 

I was hesitant at first on this PR only because I wanted to make sure serialize as a boolean made sense, vs say allowing it to be a custom function or other. But I think this is the right call, if a custom serializer is required that can be done as a transform

src/createPersistoid.js
```diff
@@ -13,8 +13,8 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
   const storageKey = `${config.keyPrefix !== undefined
     ? config.keyPrefix
     : KEY_PREFIX}${config.key}`
-
   const storage = config.storage
+  const serialize = config.serialize === false ? x => x : defaultSerialize
```
sunnylqm: This should at least return a copy of x? @rt2zz

Issue #506, opened 2017-10-26
[V5] Cannot edit reducer after persisted


hoangdoan267:
I try to add more states to my reducer
For example
Before
`const initialState = {
  recentList: [],
  historyList: [],
}`

After
`const initialState = {
  recentList: [],
  historyList: [],
  watchList: [],
  downloaded: []
}`

I found my states: watchList and downloadedList were 'undefined' though another states working normally after persisted. 
I used:
```
electron: "~1.6.2",
react: "^15.6.1",
react-redux": "^5.0.6",
redux-persist": "^5.0.1"

```
Thanks for supporting me.

rt2zz:
I am assuming you have persistReducer at the top level and are using combineReducers? The basic problem here is that you are expecting rehydration to merge 2 levels deep of your state tree, while redux-persist by default only shallow merges 1 level deep. The plan is to make this extensible via the `stateReconciler` config. Here is for example the default 1 level deep reconciler: https://github.com/rt2zz/redux-persist/blob/master/src/stateReconciler/autoMergeLevel1.js

Note most of that is just dev sanity checks, the real reconciliation is just these lines https://github.com/rt2zz/redux-persist/blob/master/src/stateReconciler/autoMergeLevel1.js#L61-L79

So ya, I think a hypothetical `autoMergeLevel2` stateReconciler would solve your use case. Question is, is there some cleaner way to expose this? e.g. this is typically the desired behavior when used with combineReducers, but not otherwise. So we could ship something like `combineAndPersistReducers`.

Open to thoughts on how to best solve this problem.

PR #511, opened 2017-10-28
(persistCombineReducers): new persistCombineReducers, recommended default for top level persistence

rt2zz:
resolves https://github.com/rt2zz/redux-persist/issues/506

v4 had a convenient behavior where it should shallow merge two levels deep, meaning additions to initialState of a reducer would automatically get included in the reconciled state. We can no longer make this assumption on v5 because persistence can be nested. 

This solves the issue by shipping a new recommended method `persistCombineReducers` which is basically redux combineReducers + setting the default state reconciler to shallow merge 2 levels deep.

In the future this may also allow us to implement some performance improvements related to how we work around the default combineReducers unrecognized state warnings. Additionally this PR adds `hardSet` state reconciler.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,16 +14,16 @@ Redux Persist is [performant](#why-redux-persist), easy to [implement](#basic-us
 ## Usage
 [API Docs](./docs/api.md)
 ```js
-import { persistStore, persistReducer } from 'redux-persist'
+import { persistStore, persistCombineReducers } from 'redux-persist'
 import storage from 'redux-persist/es/storage' // default: localStorage if web, AsyncStorage if react-native
-import rootReducer from './rootReducer'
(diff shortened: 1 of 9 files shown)


rt2zz:
If anyone has thoughts on the method name please weigh in. I chose `persistCombineReducers` for it familiarity but am open to alternatives.

micnil:
I'm so glad that I found this. Using `persistReducer(persistConfig, rootReducer)` does not work if you are using the filter transform like redux-persist-transform-filter. The initial state of the values that were not saved, were overwritten to `undefined`. 

A suggestion is to put this in to the API docs. `persistCombineReducers` is not mentioned there and it doesn't say it is the recommended way to create to root reducer. 

In any case, thank you for this awesome module!

PR #513, opened 2017-10-28
fix: localforage import

Hongbo-Miao:


rt2zz:
thanks quick catch :D

Hongbo-Miao:
Haha, thank you for the awwwwesome package!

PR #515, opened 2017-10-29
Configurable combineReducers for e.g. immutable root state

actra-gschuster:
The hardcoded combineReducers in persistCombineReducers prevents usage of e.g. redux-immutable, which makes it possible to use an immutable root state.
With the small changes in this PR one could specify through the config to use combineReducers of e.g. redux-immutable or any other custom combineReducers function.

### DIFF
--- a/src/persistCombineReducers.js
+++ b/src/persistCombineReducers.js
@@ -1,6 +1,6 @@
 // @flow
 
-import { combineReducers as baseCombineReducers } from 'redux'
+import { combineReducers } from 'redux'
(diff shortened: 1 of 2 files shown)


rt2zz:
we are going to move back to recommending `persistReducer` over `persistCombineReducers` - the latter of which was an attempt to simplify things but actually seems to have added to the confusion. 

Thanks! 

That said v5 afaik still does not have full immutable support, open to PR's to add immutable support similar to how we had it in v4 with a `stateOperators` config

PR #519, opened 2017-10-30
Allow omitting loading property in PersistGate

piotr-cz:
When using _localStorage_/ _sessionStorage_ as storage, state loading time is instant, and so there is no need to use component for loading.

Anyway IMHO using this option should be voluntary

workaround is to pass `null` as a value

### DIFF
--- a/src/integration/react.js
+++ b/src/integration/react.js
@@ -6,7 +6,7 @@ import type { Persistor } from '../types'
 type Props = {
   onBeforeLift?: Function,
   children?: Node,
-  loading: Node,
+  loading?: Node,
   persistor: Persistor,
 }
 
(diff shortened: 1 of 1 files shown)


rt2zz:
I like this, however am worried about react 15 support. Is there any way to check what version of react is being used and throw a helpful error accordingly?

piotr-cz:
Why do you think there could be problems in react v15?
It always was possible to return `null` from a `render` method, it's not something that have been introduced in v16.

On contrary returning `undefined` from render just as when the loading param is not defined throws an error `A valid React element (or null) must be returned.`

See https://codesandbox.io/s/7y91pm9q70

rt2zz:
TIL lol :D 

thanks

PR #528, opened 2017-11-01
(docs): add a diff in migration section to visualise the changes

stevemao:


stevemao:
Any comments on this one @rt2zz?

README.md
```diff
@@ -69,12 +69,53 @@ Standard Usage:
...
+     applyMiddleware(...),
+-    autoRehydrate()
+   )
+ )
+ 
+ const callback = ()
+
+ persistStore(
+   store, 
+-  config,
```
rt2zz: this middle argument needs to exist actually in order for the callback to work. It can just be a empty object `{}`, although perhaps we should update the src so as it can be an optional?

README.md
```diff
@@ -69,12 +69,53 @@ Standard Usage:
...
+  - 3. All constants (ex: `REHYDRATE`, `PURGE`) has moved to `redux-persist/lib/constants` instead of `redux-persist/constants`
+- replace `combineReducers` with **persistCombineReducers**
   - e.g. `let reducer = persistCombineReducers(config, reducers)`
 - changes to **config**:
   - `key` is now required. Can be set to anything, e.g. 'primary'
   - `storage` is now required. For default storage: `import storage from 'redux-persist/lib/storage'`
+  
+```diff
+-import { REHYDRATE, PURGE } from 'redux-persist/constants'
++import { REHYDRATE, PURGE } from 'redux-persist/lib/constants'
```
rt2zz: I am going to move constants into the root in the next release

README.md
rt2zz: actually we should do this like redux, allow it to be null

stevemao: yeah, since you mention "1. remove config argument"

README.md
stevemao: Depends on how fast you'll make a new major release I guess. We can always change this again when you make a new release 😄

rt2zz: ok its up v5.3.3, so this line can be
`import { REHYDRATE, PURGE } from 'redux-persist'`

The old format still works but this is I think preferred now.

README.md
rt2zz: this can now be null, but still needs to exist so these lines can be
```diff
-  config,
+ null,
```

README.md
```diff
@@ -69,13 +69,55 @@ Standard Usage:
 - changes to **persistStore**:
   - 1. remove config argument (or replace with an empty object)
   - 2. remove all arguments from the callback. If you need state you can call `store.getState()`
-  - 3. All constants (ex: `{REHYDRATE, PURGE}`) has moved to `redux-persist/lib/constants` instead of `redux-persist/constants`
-- replace combineReducers with **persistCombineReducers**
+  - 3. All constants (ex: `REHYDRATE`, `PURGE`) has moved to `redux-persist/lib/constants` instead of `redux-persist/constants`
```
stevemao: I can change this in a different PR.

README.md
```diff
@@ -69,13 +69,55 @@ Standard Usage:
...
+- replace `combineReducers` with **persistCombineReducers**
   - e.g. `let reducer = persistCombineReducers(config, reducers)`
 - changes to **config**:
   - `key` is now required. Can be set to anything, e.g. 'primary'
   - `storage` is now required. For default storage: `import storage from 'redux-persist/lib/storage'`
 
+```diff
+-import { REHYDRATE, PURGE } from 'redux-persist/constants'
++import { REHYDRATE, PURGE } from 'redux-persist'
```
rt2zz: lets put all of the redux-persist imports on the same line

PR #539, opened 2017-11-04
Resort to memoryStorage during tests

nnnoel:
This PR resorts to mock storage during tests, so that the `redux-persist failed to create sync storage. falling back to memory storage.` error won't occur during tests.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "flow-copy": "flow-copy-source src es && flow-copy-source src lib",
     "precommit": "lint-staged",
     "prepublishOnly": "npm run clean && npm run build",
-    "test": "BABEL_ENV=commonjs ava"
+    "test": "NODE_ENV=test BABEL_ENV=commonjs ava"
   },
   "lint-staged": {
     "src/**/*.js": [
--- a/src/storage/getStorage.js
+++ b/src/storage/getStorage.js
@@ -2,7 +2,7 @@
 
 import type { Storage } from '../types'
 
-function noop() { }
+function noop() {}
 let noopStorage = {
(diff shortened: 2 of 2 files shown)


rt2zz:
im do not believe its a fair assumption that process.env.NODE_ENV === 'test' means localstorage cannot exist. I would rather explicitly use memoryStorage in our tests, which I think is doable we just need to export memoryStorage.

nnnoel:
Yeah, I didn't mean to prevent any other test cases from using an accessible global storage type. The aim was to prevent the fallback error during tests, which occurs a lot when you're testing a big application. I could move the conditional to `process.env.NODE_ENV !== 'production' && process.env.NODE_ENV !== 'test'` before falling back on memory storage, but if you're suggesting otherwise, then feel free let me know

jozsi:
I am also facing this issue.
@rt2zz - your proposed alternative would be to use (memoryStorage from the test utils)[https://github.com/rt2zz/redux-persist/blob/master/tests/utils/memoryStorage.js] for tests?

nnnoel:
@rt2zz @jozsi should be good now 👍

rt2zz:
I chose instead to do it explicitly by using memory storage directly in the tests which should achieve the same effect 👍

jozsi:
@rt2zz - I noticed your new [storage-memory](https://github.com/rt2zz/storage-memory) package, thanks for it! I guess it worths mentioning it here.

As an alternative for those who don't want to configure a different type of storage for the tests, [jest-localstorage-mock](https://github.com/clarkbw/jest-localstorage-mock) gets the job done.

Issue #434, opened 2017-08-31
Can I upgrade to v5 on an existing app using v4 and keep my data?


williamoliveira:
I tried it but rehydration comes `undefined`

williamoliveira:
Just checked the storage now, I see the keys are different, everything is under the root key, so is there a easy way to migrate data from v4? or should I just keep on v4?

rt2zz:
there is not currently a way to migrate state from v4 to v5.

I think the way to solve this would be for redux-persist to ship with a pluggable "getStoredState":
```js
import { getStoredState } from 'redux-persist'
import { getStoredStateV4 } from 'redux-persist/lib/integration'

const v4Config = { /* whatever the old config was */ }

const config = {
  key,
  storage,
  getStoredState: async (persistConfig) => {
    let storedState = getStoredState(persistConfig)
    if (!storedState) storedState = getStoredStateV4(v4Config)
    return storedState
  }
}

persistReducer(config, baseReducer)
```

Its not the prettiest api, but the pattern would be generally useful, could be used to move from one storage engine to another for example.

**NOTE** this is not currently possible, we would need to add `getStoredState` option to config for this to work.

williamoliveira:
I dont know how the internals work, is it easy to do? do you plan to implement it?

rt2zz:
the implementation would be trivial, I just would want to take some time first to make sure this is the right api. In abstract I would like to call the config option: "migrate" or something, but unfortunately "migrate" is a loaded word, we already use that for migrating state between app state versions.

rt2zz:
I just took a first stab at it: https://github.com/rt2zz/redux-persist/tree/v5#experimental-v4-to-v5-state-migration

I have not run this code at all so it likely requires some tweaking, but it should be generally a good approach.

kelset:
Hi folks, can I revive this issue by asking if there is any update on how to migrate?

Because it looks like the migration guide that was present in the README and linked by @rt2zz earlier is now 404 😅

levsero:
@kelset https://github.com/rt2zz/redux-persist/blob/master/docs/MigrationGuide-v5.md

levsero:
@aguynamedben yep all worked as expected once the fix was in.

JennieJi:
A one time dirty hack here for using persistStore with whitelist only config:

```javascript
const v5StorageKey = `persist:${PERSIST_CONFIG.key}`;
if (!localStorage.getItem(v5StorageKey)) {
[...]

PR #546, opened 2017-11-07
fallback to default version in migrations

levsero:
Related issue/comment: https://github.com/rt2zz/redux-persist/issues/434#issuecomment-342530492

# What's changed?

Currently if you try and upgrade from v4 to v5 using https://github.com/rt2zz/redux-persist/tree/b85b56a889ebd4919c9ba11d239a4ae1861e1da6#experimental-v4-to-v5-state-migration and also add a migration, the migration will attempt to run on the v4 state which has no `_persist` version. 

This pr would allow the migration to fall back on the default version in such a scenario.

## Considerations
- I do not believe this would affect the code in any other situation as regularly this won't be executed without having the `_persist.version`.
- Is there a clearer or cleaner way to tie this code specifically to using the v4 migrate. eg. perhaps we can manually add `_persist` in the v4 `getStoredState` so the v4 migration will not affect any other parts of the code?

### DIFF
--- a/src/createMigrate.js
+++ b/src/createMigrate.js
@@ -18,8 +18,8 @@ export default function createMigrate(
         console.log('redux-persist: no inbound state, skipping migration')
       return Promise.resolve(undefined)
     }
-    // this is gauranteed to exist as version gets added before any state is stored
-    let inboundVersion: number = state._persist.version
+
+    let inboundVersion: number = (state._persist && state._persist.version) || DEFAULT_VERSION
     if (inboundVersion === currentVersion) {
       if (process.env.NODE_ENV !== 'production' && debug)
         console.log('redux-persist: versions match, noop migration')


rt2zz:
@levsero interesting, glad you are testing out the v4->v5 state migration! Is it working otherwise?

I think agree with defaulting state version here. Since the code relies on version existing, if it were to not exist for any reason (even though this *should* never happen) we want the migration to still make sense.

rt2zz:
actually I just realized the this default to default version if state version === 0 which i incorrect. Will need to update

rt2zz:
updated to include an undefined check 👍

levsero:
Good catch on the `=== 0` 👍 
Haven't come across any other issues in the v4 -> v5 migration, seems to work as intended.
💯 on all the work you've done here.

levsero:
@rt2zz wondering how soon do you plan on cutting another release? (Would like to make use of this asap in production and would prefer not having to fork anything.)

rt2zz:
I cut a release under the next tag (5.3.0-rc) with this change as well as a change to support non-object substate. I would say the changes are fairly safe so if you are able to give it a go and confirm we are 👍 I will cut 5.3.0 proper.

levsero:
@rt2zz can confirm that on 5.3.0 the migrations (both from v4-v5 as well as the state migration) all appear to be working correctly 🙇

PR #557, opened 2017-11-09
Allow migrations to be asyncronous

levsero:
# What's changed?

Was trying to add an asynchronous migration and realized that it requires a custom `createMigrate`. This pr is the edited version of the existing `createMigrate` that we're using which allows for both synchronous and asynchronous migrations.

This does use async/await which isn't used in other parts of the codebase 🤷‍♂️ 

Would love to hear any thoughts or recommendations.

## Technical description

- Adding await to the migration call causes it to always return a resolved promise and allows us to handle all migrations the same weather they are async or not.
- Using an await requires updating the wrapping function to be async, this also forces all responses to be Promise.resolves so no longer have to explicitly return `Promise.resolve`s.
- Don't need to have the last try catch as it will automatically return a `Promise.reject` in the case of an error.

### DIFF
--- a/src/createMigrate.js
+++ b/src/createMigrate.js
@@ -9,14 +9,14 @@ export default function createMigrate(
   config?: { debug: boolean }
 ) {
   let { debug } = config || {}
-  return function(
+  return async function(
     state: PersistedState,
     currentVersion: number
   ): Promise<PersistedState> {
     if (!state) {
       if (process.env.NODE_ENV !== 'production' && debug)
         console.log('redux-persist: no inbound state, skipping migration')
-      return Promise.resolve(undefined)
+      return undefined
     }
 
     let inboundVersion: number =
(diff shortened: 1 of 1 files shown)


rt2zz:
I am +1 on this conceptually. 

I do however think it is probably to early to introduce async await. Open to discussion on the topic, but IIRC that basically means users will require the regenerator runtime which I think is an unfair burden. I recall we momentarily added async in the default storage engine which caused some consternation.

If anyone feels that async await should be standard now please chime in - I would love to start using it in the codebase.

levsero:
@rt2zz I don't have the most context on how the babel/js build process works so I may be missing something. 
But the `regenerator-runtime` is already included in the package dependencies (`package-lock.json`) and running `npm run build` locally does properly compile the `async`/`await` to the lib and es directories with `_asyncToGenerator`, so I'm not totally clear on what this would be adding to the project?

levsero:
Closing until such time can use async await.

PR #559, opened 2017-11-13
fix numerical sort

levsero:
Need to add a sort function for javascript to correctly sort numbers. The version keys are numbers so would not currently be sorting correctly.


rt2zz:
wow, never knew :o

tyty

Issue #568, opened 2017-11-15
'File' or 'Blob' getting rehydrated as empty objects 


alexgurr:
I have various files stored in my reducer. When I reload the page, the files are getting rehydrated as empty objects?

I'm using localForage, which supposedly supports the storing of blobs and files.

Any ideas?

rt2zz:
right the problem here is that redux-persist calls stringify which does not support blobs (afaik). We should probably have a way to turn off serialization - I think that would likely solve your case.

alexgurr:
Any chance we could achieve this soon? I'm happy to help. I'm just thinking about current project deadlines and whether I need to start looking for alternatives :)

reblws:
I also made a PR regarding this in #507

alexgurr:
Good shout. My implementation supports custom serialise/deserialise functions though (including turning them off), rather than just on or off.

reblws:
I considered that before but decided against it because it seemed like it would be duplicating the functionality of the `transforms` argument.

alexgurr:
Serialising is not the same as transforming though. Transforming allows us to manipulate state before storage, whereas serialising is determining the format of the data in state.

rt2zz:
Open to others weighing in here - I prefer the minimal change which is serialize as boolean, but if there are many valid use cases for custom serialize then it may make sense to discuss an alternative implementation. 

If we do go that route I would like to factor in how this my interplay with async transforms which has been a long standing request from many.

PR #570, opened 2017-11-15
Expose serialize/deserialize as config options

alexgurr:
As a result of https://github.com/rt2zz/redux-persist/issues/568


alexgurr:
Exposing serialize/deserialize allows us to pass an empty function and turn them off effectively.

PR #572, opened 2017-11-16
Expose serialize/deserialize as config options

alexgurr:
Exposing serialize/deserialize allows us to pass an empty function and turn them off effectively.


rt2zz:
I ended up going with https://github.com/rt2zz/redux-persist/pull/507/ which has the same api but a slightly different implementation, thanks for spurring action here 🎈

rt2zz:
ah I just saw about allowing customer serialize / deserialize. I am pretty conservative with changes at this point so I would prefer to go with the simple serialize as boolean option for now. If the need for this grows we can continue the discussion

PR #587, opened 2017-11-25
Marked inbound and outbound parameters on createTransform as optional

wvteijlingen:
Since line 26 and 27 check whether they are set, it seems logical to me that the parameters should allow null/undefined.


rt2zz:
👍

PR #588, opened 2017-11-27
Updating readme docs.

thenewt15:
This is a re-write of several sections in the readme. Added a lot of stuff I wish I knew when I was getting started with the library. 

The diff might be hard to sort through because there are so many changes...easiest way to read it may be to copy the markdown into a viewer like [dillinger.io](https://dillinger.io/).

- Created a "Quickstart" section. Tried to keep it super simple and discuss more advanced customization after that section. The Quickstart section uses `persistReducer` instead of `persistCombineReducers`. Since most people will come into here already having redux set up (with combineReducers already set up), I thought perhaps we could lead people toward `persistCombineReducers` when discussing more advanced customizations. That way we aren't asking them to refactor combineReducers in the QuickStart.
- I elaborated a little more on nested persistors and think this will help people take better advantage of this feature.
- Tweaked the the section on transforms and included a new example.
- I moved all the details of v4 to v5 migration to a separate file. I've seen a lot of other libraries do this where they have a standalone Migration Guide. I didn't edit the contents of this migration guide. Just a copy/paste from stuff in the current readme. 
- I didn't get around to editing the section on migrations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,169 +1,158 @@
 # Redux Persist
+
 Persist and rehydrate a redux store.
 
-Redux Persist is [performant](#why-redux-persist), easy to [implement](#basic-usage), and easy to [extend](./docs/ecosystem.md).
+[![build status](https://img.shields.io/travis/rt2zz/redux-persist/master.svg?style=flat-square)](https://travis-ci.org/rt2zz/redux-persist) [![npm version](https://img.shields.io/npm/v/redux-persist.svg?style=flat-square)](https://www.npmjs.com/package/redux-persist) [![npm downloads](https://img.shields.io/npm/dm/redux-persist.svg?style=flat-square)](https://www.npmjs.com/package/redux-persist)
 
-`npm i redux-persist`
+Redux Persist takes your redux state object and saves it to persisted storage. On app launch, it retrieves this persisted state and saves it back to redux.
 
-[![build status](https://img.shields.io/travis/rt2zz/redux-persist/master.svg?style=flat-square)](https://travis-ci.org/rt2zz/redux-persist)
-[![npm version](https://img.shields.io/npm/v/redux-persist.svg?style=flat-square)](https://www.npmjs.com/package/redux-persist)
-[![npm downloads](https://img.shields.io/npm/dm/redux-persist.svg?style=flat-square)](https://www.npmjs.com/package/redux-persist)
+**Note:** These instructions are for redux-persist v5. For a list of breaking changes between v4 and v5, see our [migration guide](./docs/MigrationGuide-v5.md).
+[v4](https://github.com/rt2zz/redux-persist/tree/v4.8.2) will be supported for the forseeable future, and if it works well for your use case you are encouraged to stay on v4.
 
-**Note:** These docs apply to redux-persist v5. [v4](https://github.com/rt2zz/redux-persist/tree/v4.8.2) will be supported for the forseeable future, and if it works well for your use case you are encouraged to stay on v4.
+## Quickstart
+##### Install package
+`npm install --save redux-persist`
+\- OR -
+`yarn add redux-persist`
 
-## Usage
-[API Docs](./docs/api.md)
+##### Implementation
+When creating your redux store, pass your [createStore](https://redux.js.org/docs/api/createStore.html) function a PersistReducer that wraps your app's root reducer.
+Once your store is created, pass it to the `persistStore` function, which ensures your redux state is saved to persisted storage whenever it changes.
+
 ```js
-import { persistStore, persistCombineReducers } from 'redux-persist'
-import storage from 'redux-persist/es/storage' // default: localStorage if web, AsyncStorage if react-native
-import reducers from './reducers' // where reducers is an object of reducers
+import { createStore } from 'redux';
+import { persistStore, persistReducer } from 'redux-persist';
+import storage from 'redux-persist/lib/storage';
 
-const config = {
+import rootReducer from './reducers'; // the value from combineReducers
+
+const persistConfig = {
   key: 'root',
-  storage,
-}
+  storage: storage
+};
 
-const reducer = persistCombineReducers(config, reducers)
-
-function configureStore () {
-  // ...
-  let store = createStore(reducer)
-  let persistor = persistStore(store)
-
-  return { persistor, store }
-}
+const persistReducer = persistReducer(persistConfig, rootReducer);
+export const store = createStore(persistReducer);
+export const persistor = persistStore(store);
 ```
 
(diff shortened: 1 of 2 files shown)


rt2zz:
thanks for doing this, definitely going to take a lot of these changes. The one that I am uncertain about is recommending persistReducer over persistCombineReducers. The main difference therein is the merge strategy, persistCombineReducers applies two levels of shallow merge whereas persistReducer only does 1 level shallow merge. Need to think on this a bit longer..

thenewt15:
Would using persistReducer and with this config result in the exact same thing as using persistCombineReducers?

```
const persistConfig = {
  key: 'root',
  storage: storage,
  stateReconciler: autoMergeLevel2
}
```

Is there a benefit to using persistCombineReducers other than its merge level?

I'm trying to sort out why persistCombineReducers exists at all. It seems to do two things:
1. Defaults stateReconciler to autoMergeLevel2, which you could set explicitly with persistReducer.
2. Calls combineReducers for you. Is there a benefit to this though? People will likely already have combineReducers set up when they start implementing this library. What do they gain by passing all their reducers into persistCombineReducers rather than creating this root reducer themselves with combineReducers?

thenewt15:
@rt2zz a follow-up question for you...

I'm trying to finalize my data store on my current project and ultimately just trying to figure out whether I should be using autoMergeLevel2 for my top-level PersistReducer (i.e. the PersistReducer that wraps my root reducer returned from combineReducers).

I noticed that the persistCombineReducers function does this for you (uses autoMergeLevel2), but I can't wrap my head around why this is needed. autoMergeLevel1 looks like it makes a copy of state, which should successfully notify all my container components via mapStateToProps. What additional benefit would I get by having copies made two levels deep?

rt2zz:
@thenewt15 correct, persistCombineReducers is really no more than defaulting to autoMergeLevel2 and applying combineReducers. Those can be done separately if desired. We put them together in one method actually in an attempt to reduce confusion, since "autoMergeLevelX" is not a very user friendly name. Generally speaking the intent of autoMergeLevel2 is to work at the top level, whereas autoMergeLevel1 is intended for use when nested within your reducer tree.

The only difference between the two reconcilers is how many levels of state they shallow merge. If you want initialState merged in with incoming rehydrated state then you need level 2, otherwise level 1 is perfect.

Sorry for confusion and slow response. I know right now stateReconciler is the biggest pain point around understanding and implementing redux persist. I am looking for ways to improve the api around this!

tldr; yes using persistReducer (which is default autoMergeLevel1) is quite fine, just be aware that your sub reducer initial state's will not be shallow merged in during rehydrate.

thenewt15:
@rt2zz I can add something to my PR on how merging works. I'd like to clarify a couple things about it though. Are these statements true?

1. Once the persisted state is retrieved from storage on app launch, redux-persist dispatches a REHYDRATE action. After the REHYDRATE action is sent to all your reducers, the merging process begins. 
2. The term "rehydrated state" refers to the situation where you catch the REHYDRATE action in a reducer and update the state object. The state returned from this REHYDRATE case is the "rehydrated state."
3. The merging process checks whether you changed the state object when the REHYDRATE action was dispatched (i.e. whether the rehydrated state is equal to the initial redux state). If it was changed, then the initial redux state is updated with this rehydrated state. If it was not changed, the initial redux state is updated with whatever was persisted (i.e. inboundState). 
4. redux-persist v5 introduced auto-merging, which means you no longer need to catch the REHYDRATE action within your reducers. Without catching this action, the initial redux state will simply be updated with any persisted state.


Ok, so regarding the two merge levels, I'm still confused about when you'd want to use autoMergeLevel2. Let's say I have a state object like this.
```
{
  auth: { currentUser: { firstName: 'Sally' }, isLoggingIn: false },
  otherStateKey: { ... }
}
```
With autoMergeLevel1, I've noticed that my restored redux state will include any changes I make to `isLoggingIn`, or even to `currentUser.firstName`, if I update these from the REHYDRATE action in my Auth Reducer.
```
  case (REHYDRATE): 
    return {
      ...state,
      isLoggingIn: true,
      currentUser: { ...state.currentUser, firstName: 'Bob' }
    };
```
Because I'm always returning a copy of state (like you always should), it doesn't matter how deep or shallow these changes go. I actually haven't been able to find a case where autoMergeLevel1 didn't pick up a change I made to state. Can you provide an example of a state change that wouldn't be detected?

levsero:
@thenewt15 how would it work if you don't want to manually listen for `case REHYDRATE` in all your reducers? I assume this is what the deep merge is for, because the rehydrated state is not going to `===` the default state, so you need to deep merge the keys to merge them.

thenewt15:
thank you @levsero, I misunderstood what the merge process was trying to handle. After looking at the code, I'm seeing that its meant to handle changes to your initial redux state. So if your auth state had its initial state set like this.
```
const INITIAL_STATE = {
  currentUser: null,
  isLoggingIn: false,
};
```
And then you released a new version of your app that set it like this.
```
const INITIAL_STATE = {
  currentUser: null,
  isLoggingIn: false,
  error: ''
};
```
You'd obviously want your reconciled state object to include this new error key. But `autoMergeLevel1` would cause it to not be included, because it would just override this initial state with whatever was persisted.
`autoMergeLevel2` overcomes this problem, because it would first make a copy of the initial state, and then only override the keys within the auth object that were persisted. Since 'error' wouldn't be persisted, it would be left alone.

rvirani1:
Just a note, I've had two developers on my team struggle with implementing v5. It was only after we found this PR did we really get a sense of how v5 worked. The current docs on the main README are still inadequate in our opinion. Would love for a version of this PR to get merged in.

thenewt15:
Thanks @rvirani1 I may move this to a blog post if it remains here getting no love. There's actually a lot I'd like to clean up with the docs, but I prefer to see outstanding PRs merged in before I submit new ones.

rvirani1:
A blog post would be great.

ScreamZ:
Nice explanations, really loved that.

It could go a little deep in a concept though, like the **key** that can be used to make a "Vary" on userId per example.

Also detailing each config props a bit further but also especially the api https://github.com/rt2zz/redux-persist/blob/master/docs/api.md : stateReconciler, throttle, keyPrefix, ...

Thanks anyway.

Spread the love <3

thenewt15:
@rvirani1 @ScreamZ I put this on medium, with an additional writeup on the merge process. Still hoping this gets merged at some point, but in the meantime I think this will help some people.

https://medium.com/@thenewt15/the-definitive-guide-to-redux-persist-84738167975

rt2zz:
going to get this in now, will do a second revision pass after

rt2zz:
thanks @thenewt15 !

Issue #601, opened 2017-12-04
Hot module replacement using _persisst


ScreamZ:
Hey,

I'm implementing HMR in my app using the following code 

```js
import { combineReducers } from 'redux'
import configureStore from './configureStore'

/* ------------- Assemble The reducers ------------- */
export const reducers = combineReducers({
  nav: require('./reducers/navigation').reducer,
  user: require('./reducers/user').reducer
})

export default () => {
  const storeTools = configureStore(reducers)

  if (module.hot) {
    module.hot.accept(() => {
      const nextRootReducer = require('./').reducers
      storeTools.store.replaceReducer(nextRootReducer)
    })
  }

  return storeTools
}
```

It is supposed to work well, but the app display the following message, that is annoying for on a HMR purpose with message the following message:

`Unexpected key "_persist" found in preloadedState argument passed to createStore. Expected to find one of the kwown reducer keys instead "nav", "user". Unexpected keys will be ignored.`

Any idea how can I deal with that ?

Best regards

rt2zz:
Does it still warn if you use persistCombineReducers instead of combineReducers?

ScreamZ:
@rt2zz Seems to work boss ! 

I have to adapt my code, I can give you an example with a PR for the documentation, I can see this part is "Coming soon" are you interested in ?

ScreamZ:
I'm encountering new issues, can we talk about it in the above PR ? Thanks

blackLearning:
For me, persist reducer in replaceReducer solves the problem.

```
export const store = createStore(
  persistReducer(config, rootReducer),
  compose(applyMiddleware(...middlewares), ...enhancers)
);

if (module.hot) {
  module.hot.accept(() => {
    const nextRootReducer = require('../reducer/index').default; // eslint-disable-line
    store.replaceReducer(persistReducer(config, nextRootReducer))
  })
}
```

ScreamZ:
It should works, true :) Bug it could be nice to use the built-in method (see my PR #607)

rt2zz:
@ScreamZ thanks, I modified your PR to be a little more inline with standard setup / simplify.

I apologize I lost the attribution on the squashed commit.

PR #607, opened 2017-12-07
Update readme to present HRM pattern

ScreamZ:
As I had the issue as seen here https://github.com/rt2zz/redux-persist/issues/601, I guess this was a nice idea to share my setup with the community.

I present my pattern to handle HMR with `redux` and `redux-persist`.

**EDIT** : It should work for most people but those using Immutable structure ('seamless-immutable' in my case) might encounter an issue with the implementation of persistCombineReducers & Immutable structure using this pattern.

For instance I'm trying to use this now because _persist reducer is now passing through the transformers... This is not the cause with combineReducer method. Even though this is not solving my problem.

```js
const immutableTransformerState = {
  out: state => {
    if (state.hasOwnProperty('version') && state.hasOwnProperty('rehydrated')) {
      return state
    }
    return Immutable(state)
  },
  in: raw => {
    if (Immutable.isImmutable(raw)) {
      return raw.asMutable({ deep: true })
    }

    return raw
  }
}
```

I'm still getting some reducer are not Immutable but they should.
It looks like reducer are processed by the transformers but something is destroying my Immutable from transformers and override them with a standard object... any idea ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ class App extends Component {
 Additional Usage Examples:
 1. [Nested Persists](#nested-persists)
 3. Code Splitting [coming soon]
-4. Hot Module Reloading [coming soon]
+4. [Hot Module Replacement](#hot-module-replacement)
 
 ## v5 Breaking Changes
 There are three important breaking changes.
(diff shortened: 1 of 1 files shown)


blackLearning:
It seems you're still tring to use the original reducer to replace the persisted reducer, you sure this will work for HMR? 

I'm using the seamless-immutable too, here is my setup, it works fine to me.

```
const myTransform = createTransform(
  // transform state coming from redux on its way to being serialized and stored
  state => (Immutable.isImmutable(state) ? Immutable.asMutable(state) : state),
  // transform state coming from storage, on its way to be rehydrated into redux
  state => Immutable(state)
)

const config = {
  key: 'root',
  debug: true,
  blacklist: ['nav', '_persist'],
  transforms: [myTransform],
  storage,
};

export const store = createStore(
  persistReducer(config, rootReducer),
  compose(applyMiddleware(...middlewares), ...enhancers)
);

store.runSaga = sagaMiddleware.run;
store.close = () => store.dispatch(END);

export const persistor = persistStore(store, null);

// clear storage when app is reloaded.
// if (__DEV__) persistor.purge();


if (module.hot) {
  module.hot.accept(() => {
    const nextRootReducer = require('../reducer/index').default; // eslint-disable-line
    store.replaceReducer(persistReducer(config, nextRootReducer))
  })
}

```

ScreamZ:
I'll try your solution and will update my PR accordingly if it works for me :)

Thanks for the explanation :)

ScreamZ:
@blackLearning @rt2zz 

Seems good now, I updated readme accordingly to @blackLearning proposal.

Best regards

rt2zz:
incorporated, thanks!

PR #610, opened 2017-12-09
Avoid creating new state each time, trigger repeat render.

helele90:
if we have not already updated state, straight passthrough.
Avoid creating new state each time, trigger repeat render.

### DIFF
--- a/src/persistReducer.js
+++ b/src/persistReducer.js
@@ -136,13 +136,18 @@ export default function persistReducer<State: Object, Action: Object>(
       }
     }
 
+    const reducedState = baseReducer(restState, action)
+
     // if we have not already handled PERSIST, straight passthrough
-    if (!_persist) return baseReducer(state, action)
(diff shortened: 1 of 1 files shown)


src/persistReducer.js
```diff
@@ -139,6 +139,10 @@ export default function persistReducer<State: Object, Action: Object>(
     // if we have not already handled PERSIST, straight passthrough
     if (!_persist) return baseReducer(state, action)
 
+    // if we have not already updated state, straight passthrough
+    const reducedState = baseReducer(restState, action)
+    if (reducedState === state) return reducedState
```
rt2zz: I think this should be `if (reducedState === restState) return state`

rt2zz:
crap I forgot about this PR and reimplemented it: https://github.com/rt2zz/redux-persist/pull/689

thanks nevertheless!

PR #615, opened 2017-12-12
(type definitions) Included typescript definition file with the minim…

rodrigoelp:
Contains the minimum type definition to get it to work in react-native with the `PersistGate` component.

This is a few pull request to get this work in (unblocking people needing the type definition) but will continue expanding on this definition as I bump into other scenarios (I need ways to test functionality as I am not an expert on reactjs)

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,0 +1,131 @@
+declare module "redux-persist" {
+    import { Reducer, ReducersMapObject } from "redux";
+
+    export type PersistState = {
+        version: number,
+        rehydrated: boolean
+    }
+
+    export type PersistedState = {
+        _persist: PersistState,
+    } | void
+
+    export type Transform = {
+        in: (state: Object, key: string) => Object,
+        out: (state: Object, key: string) => Object,
+        config?: PersistConfig
+    }
+
+    export type PersistConfig = {
+        version?: number,
+        storage: Object,
+        key: string,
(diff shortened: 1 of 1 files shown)


rt2zz:
awesome, will getting this in break anything for existing TS users that may have shimmed this or otherwise?

If not I prefer to merge this asap and get as many eyes on it as possible. Putting this on master now, but will wait before release.

rt2zz:
thank you 💥

rodrigoelp:
Hi @rt2zz, master (v5) is broken for any ts user... this will allow people to use typescript.

rodrigoelp:
By the way @rt2zz, which other area is pressing to get in the type definition? I basically exposed what's needed to use `PersistGate`, should I get the `redux-persist/es/storage` module next?

If so, do you know how to test it with react-native? If you don't, do you have a set of steps I can follow to test it in my dev env (web)?

rt2zz:
yes I think storage is the other big one. Of course if one uses localForage or AsyncStorage then this is moot as these work out the box (and I am assuming both have accessible TS defs?)

I do not believe this requires react-native specific testing, as the required storage api is identical for all environments.

Beyond that there are the stateReconcilers (https://github.com/rt2zz/redux-persist/tree/master/src/stateReconciler) but very few people will consume these directly so are much less urgent.

PR #616, opened 2017-12-12
Remove deprecated v4 typescript declaration file & add typings for v5

ethan605:
# Checklist

* [x] Constants
* [x] `persistReducer`
* [x] `persistCombineReducers`
* [x] `persistStore`
* [x] `createMigrate`
* [x] `createTransform`
* [x] `getStoredState`
* [x] `createPersistoid`
* [x] `purgeStoredState`
* [ ] Storage functions
* [ ] State reconciler functions
* [ ] Integration functions


rt2zz:
woo nice 

I like having them split out per file, will make maintenance easier I think, and help when comparing with flow defs.

cc @rodrigoelp

ethan605:
@rt2zz yeah except simple functions (like `createPersistoid`, `getStoredState`, `purgeStoredState` are placed in `index.d.ts`), another are split into individual files which are reflected with current flow files

rodrigoelp:
Hi, I will get it to a point it works, we can then split it into multiple files. I know is messy like this but I am trying to get my head around the library as well as producing the type definition

rodrigoelp:
Hi @rt2zz, do you have examples that I can use to test the integration with typescript?

I've been creating the type definitions for these and I want to be able to test some of the other functions that I have not used before.

inakianduaga:
@rodrigoelp 

I think the signature of `persistCombineReducers` should be

```ts
  /**
   * It provides a way of combining the reducers, replacing redux's @see combineReducers
   */
  export function persistCombineReducers<S>(
    config: PersistConfig,
    reducers: ReducersMapObject
  ): Reducer<S & PersistedState>
```
since the real root state will have the added keys from `_persist`? 

We should also add a 
```ts
declare module 'redux-persist/lib/storage'
```

so typescript is happy when the ES5 version gets imported (my Jest config doesn't like this particular import)

rodrigoelp:
Hi @inakianduaga,

Thanks for the comment, I will definitely include it in the next commit as I have been trying to create the type definition as separate files as mentioned by @ethan605. Haven't finished the definition because I have hit a wall doing that activity as I am sort of new to typescript and have been learning this part of the code generation.

I got the storage part already done (although the module I got exposed is `redux-persist/es/storage` instead.

LKay:
I totally missed that similar PR was already up. I also made some improvements to the current types definitions. Maybe something will be useful to add/modify in this PR? https://github.com/LKay/redux-persist/blob/lkay/typings/src/index.d.ts

src/typings/createTransform.d.ts
```diff
@@ -0,0 +1,17 @@
+declare module 'redux-persist/createTransform' {
+  type TransformConfig = {
+    whitelist?: string[],
+    blacklist?: string[],
+  }
+
+  type Transformer = {
```
LKay: Id rather define this as generic `Transform<State, Raw>` the same way as it was in the v4. At least we won't break types for depending libraries such as `redux-persist-transform-filter` etc.

rodrigoelp: Hi @LKay, I have created a pull request today which has the entire definition for `redux-persist` though is not split into multiple files.

In this pull request, the transformer is a generic definition, do check it here: #624

inakianduaga: @LKay 
```ts
   export function persistCombinedReducers<State>(config: PersistConfig, reducers: ReducersMapObject): Reducer<State>;
```

should include the state coming from redux_persist itself, so it should be

```ts
   export function persistCombinedReducers<State>(config: PersistConfig, reducers: ReducersMapObject): Reducer<State & PersistedState>;
```

rt2zz:
I went ahead and merged https://github.com/rt2zz/redux-persist/pull/624 as it looks good to me and @rodrigoelp did the initial implementation.

That said, I am not a typescript user and would love to continue to improve the types. Any TS users please weigh in on best way to move forward. Is it idiomatic to keep all types in one file, or split out into separate?

LKay:
@rt2zz You can go both ways, either will work. More clean version, imo would be to store types definitions the same way as you did with flow types in separate files as it's better to maintain but i don't mind having them in one file. I also saw some things that should be fixed or be of more proper type. Will make some changes and bring up PR on Monday.

rt2zz:
we merged the follow up PR. If anyone feels strongly a reorg is needed please reopen or comment. Thanks!

PR #620, opened 2017-12-13
export function defaultDeserialize for v5 crosstab

TSMMark:
See https://github.com/rt2zz/redux-persist-crosstab/pull/17#issuecomment-351456003


TSMMark:
Any problems with this or can we get a merge? 😄

PR #621, opened 2017-12-13
Improve autoMergeLevel2

EricPKerr:
This fixes an issue where the persisted state of a reducer changes shape from the new initialState of that reducer.  The current implementation leaves all keys in place from both versions instead of maintaining the new state shape.  This could result in duplicate data being stored, making it tougher to track down crashes and bugs.

Currently:

```
let inboundState = {
  reducerA: {
    key1: true,
    key2: [1, 2, 3]
  }
}

let reducedState = {
  reducerA: {
    key2: [],
    key3: true
  }
}

// Merged newState
newState = {
  reducerA: {
    key1: true,
    key2: [1, 2, 3],
    key3: true
  }
}
```

It should be:
```
newState = {
  reducerA: {
    key2: [1, 2, 3],
    key3: true
  }
}
```

### DIFF
--- a/src/stateReconciler/autoMergeLevel2.js
+++ b/src/stateReconciler/autoMergeLevel2.js
@@ -35,22 +35,8 @@ export default function autoMergeLevel2<State: Object>(
         Object.keys(inboundState[key]).forEach(prop => {
           // if the newState doesn't have the stored prop, skip it
           if(!newState[key].hasOwnProperty(prop)) return;
-
-          const newPropType = typeof newState[key][prop];
-          const inboundPropType = typeof inboundState[key][prop];
-
-          // if the inboundState has the same Level2 prop and it's the same type then merge it. Otherwise skip it.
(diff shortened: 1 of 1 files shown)


rt2zz:
hm, this could go both ways. I like this pattern but it basically means users are required to maintain accurate initialState in their reducer which is an opinionated change.

I have been seeing over time most of the opinionated pieces in redux-persist are in the stateReconciler, so I am wondering what we might be able to do to accommodate these needs without allowing complexity to explode.

EricPKerr:
@rt2zz Totally understand the opinionated change aspect of this and wanting to avoid that if possible.  Are there any scenarios where you'd want keys maintained that aren't in the `initialState`?  Presumably the rest of your app codebase would be referencing that, but it's the persisted older `reducedState` that's not correctly synced.

Open to discussion though - where are there more opinonated pieces in the stateReconciler?

PR #624, opened 2017-12-15
(types) Extending type definitions to all modules in redux-persist

rodrigoelp:
Created the type definition for every function, interface and instance.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,108 +1,211 @@
+
 declare module "redux-persist" {
-    import { Reducer, ReducersMapObject } from "redux";
+    export * from "redux-persist/es/constants";
+    export * from "redux-persist/es/types";
+    export * from "redux-persist/es/createMigrate";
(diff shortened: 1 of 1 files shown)


rodrigoelp:
Hi guys, In this pull request you will find a single file with all the type definitions segregated by modules. I still haven't pulled things on separate files as I had issues doing that (and the default exports for local storage, session and native storage gave me issues as well, I have not tested any of these yet)

I have not tested every declaration as I do not have an example for migration or state reconcilers.

It does not have any breaking in comparison to the previous commit.

rt2zz:
awesome! thank you!

rodrigoelp:
No worries :) glad I've been useful :)

rodrigoelp:
By the way, should we update `Readme.md` file to remove the comment `Additionally v5 does not yet have typescript bindings.`?

rt2zz:
done

rodrigoelp:
Great, you are quick ;)

PR #629, opened 2017-12-18
catch migration error when using filestorage on android

StefanWegener:
When updating from v4 to v5 I have an issue when I use redux-persist-filesystem-storage on Android.

The framework throw an FileNotFoundError, thats why the getStoredStateV4 function gets never called.

Catch the error and call the getStoredStateV4 work's for me.

### DIFF
--- a/src/integration/getStoredStateMigrateV4.js
+++ b/src/integration/getStoredStateMigrateV4.js
@@ -17,7 +17,10 @@ export default function getStoredState(v4Config: V4Config) {
     return getStoredStateV5(v5Config).then(state => {
       if (state) return state
       else return getStoredStateV4(v4Config)
-    })
+    }).catch(e => {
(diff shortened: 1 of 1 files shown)


levsero:
Are you using the v4->v5 migration as described in https://github.com/rt2zz/redux-persist#experimental-v4-to-v5-state-migration? I assume `redux-persist-filesystem-storage` is still supported in v5 so don't see why it should cause an error.  

(Should this be opened as an issue first/additionally for discussion about what the cause is?)

rt2zz:
agreed, we need to understand why it is failing for v5 otherwise this will lead to unexpected behavior. If the framework throws a FileNotFoundError anytime we try to access a non-existent key, then the storage engine will need to be updated to not rethrow that error and instead just return undefined state.

PR #642, opened 2017-12-28
Add comment about how to import commonJS

maggiepint:
We have some old webpack and this bit me for a few minutes. Thought I would add a comment for the next person.


rt2zz:
I went ahead and made the docs always use lib: https://github.com/rt2zz/redux-persist/commit/d238b7f02b4e55eee198e518c86f8aa7e8384104 since this seems to be a common hangup.

in the near future I would like to allow importing without lib/es so something like `import storage from 'redux-persist/storage'` then the module system can decide for itself lib vs es

PR #662, opened 2018-01-17
Implicitly include _persist key in whitelist.

thenewt15:
We should include the _persist key in the whitelist, even if the user didn't explicitly include it. 

I discovered this bug when I noticed that my migration was running every time I Iaunched my app. This PR should fix that issue.


rt2zz:
thank you, will cut a new release soon

Issue #669, opened 2018-01-19
React Native app crashing immediately after adding redux-persist


axelkennedal:
I had redux and react-redux working perfectly with my app and wanted to add persistence, so I followed the guide in the readme of this repo, but when I attempt to launch my app it just crashes without any helpful feedback.

I've described the problem in detail [here](https://stackoverflow.com/questions/48344836/redux-persist-crashing-immediately).

rt2zz:
I just updated the docs to be `loading={null}` with a note below about usage

PR #670, opened 2018-01-19
Wrap "Loading.." component on readme

eduduardo:
Resolve issue #669 

When running the loading with plain text causes an exception:
"Cannot Add a child that doesn't have a YogaNode to a parent with out a measure function"

And crashes the app without report. Just wrap the loading or create another Loading component.


rt2zz:
@eduduardo thanks! I actually just updated the docs to set loading to null. Trying to keep the initial usage as simple as possible (also react-native agnostic). I think null is the best starting point for people to get up and running without worrying about non-essential logic.

eduduardo:
@rt2zz cool!

PR #671, opened 2018-01-20
fix(typescript): Specify path for typescript definitions

iRoachie:
The typescript definitions weren't being picked up previously. This is because the main file for the package is in the `lib` directory so the editors were looking for an `index.d.ts` file in the `lib` folder. 

This PR specifies the path where the typings are located.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "5.5.0",
   "description": "persist and rehydrate redux stores",
   "main": "lib/index.js",
+  "types": "src/index.d.ts",
   "module": "es/index.js",
   "repository": "rt2zz/redux-persist",
   "files": [


iRoachie:
Closing, just saw message regarding v5.5 not yet having typescript definitions

rt2zz:
we not have v5 TS definitions, and included the reference in package.json 

thanks!

iRoachie:
Sweet! One question, if I update to v5 will the persisted store on the client be affected in anyway?

rt2zz:
yes, unfortunately the state is stored in a different format so you have two options:
1. do nothing and all users will effectively have their state reset
2. use a v4 -> v5 migration. A rough explanation of how that works is here: https://github.com/rt2zz/redux-persist/blob/master/docs/MigrationGuide-v5.md#experimental-v4-to-v5-state-migration - I know at least a few users have implemented this successfully but YMMV

iRoachie:
Ahh :\, alrighty then I'll hold it off to a next major release for my app. Thanks

Issue #675, opened 2018-01-22
Type mismatch: Property '_persist' is missing in type 'State'.


openscript:
The typing for `PersistedState` is not accurate in my opinion, which leads to this error:

```
ERROR in [at-loader] ./src/index.tsx:36:5 
    TS2345: Argument of type 'Reducer<(State & void) | (State & { _persist: PersistState; })>' is not assignable to parameter of type 'Reducer<State>'.
  Types of parameters 'state' and 'state' are incompatible.
    Type 'State' is not assignable to type '(State & void) | (State & { _persist: PersistState; })'.
      Type 'State' is not assignable to type 'State & { _persist: PersistState; }'.
        Type 'State' is not assignable to type '{ _persist: PersistState; }'.
          Property '_persist' is missing in type 'State'.
```

```
[...]
    export interface PersistState { version: number; rehydrated: boolean; }
    export type PersistedState = { _persist: PersistState } | void;

    export interface PersistConfig {
[...]
```

I will create a pull-request.

openscript:
Was fixed with #725.

rt2zz:
just published v5.9.1 👍 ty

CyxouD:
Reproduced with `redux-persist`: `6.0.0`. @rt2zz could this bug appear again?

Fixed by wrapping `State` type in `CombinedState` of `redux`

![image](https://user-images.githubusercontent.com/17162008/88268225-c77cc080-ccda-11ea-9e59-e39e9c0ef6d2.png)


React native info:
```
System:
    OS: macOS Mojave 10.14.5
    CPU: (8) x64 Intel(R) Core(TM) i7-4850HQ CPU @ 2.30GHz
    Memory: 387.30 MB / 16.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 10.17.0 - /usr/local/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.11.3 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.8.4 - /usr/local/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.0, DriverKit 19.0, macOS 10.15, tvOS 13.0, watchOS 6.0
    Android SDK: Not Found
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.5900203
[...]

Issue #674, opened 2018-01-22
package.json doesn't refer to types


openscript:
The `package.json` should reference the types.

```
[...]
  },
  "types": "src/index.d.ts",
  "devDependencies": {
[...]
```

I'll create a pull-request for this.

BeaveArony:
I think it should be "typings" instead of "types". At least that's what is working for me.

openscript:
According to https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html it should be `types`.

PR #676, opened 2018-01-22
Improve Typescript typing according to #674 and #675

openscript:
Reference TypeScript types according to: https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html
Using an interface over a type. (see https://stackoverflow.com/questions/37233735/typescript-interfaces-vs-types)
Using optional properties instead of void.
Cleaned up the types with "tslint:recommended".

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -1,77 +1,76 @@
-
-declare module "redux-persist" {
-    export * from "redux-persist/es/constants";
-    export * from "redux-persist/es/types";
-    export * from "redux-persist/es/createMigrate";
-    export * from "redux-persist/es/createPersistoid";
-    export * from "redux-persist/es/createTransform";
-    export * from "redux-persist/es/createWebStorage";
-    export * from "redux-persist/es/getStoredState";
-    export * from "redux-persist/es/persistCombineReducers";
-    export * from "redux-persist/es/persistReducer";
-    export * from "redux-persist/es/persistStore";
-    export * from "redux-persist/es/purgeStoredState";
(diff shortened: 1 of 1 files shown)


rt2zz:
cc @LKay @rodrigoelp

these changes look good to me but Ill defer to TS users

LKay:
I don't see the point of replacing double quotes with single quotes which is 99% of all changes here. Basically there is only one line that matters here.

openscript:
I can revert these if you want, but I suggest to have something like `tslint`, to keep the code clean and organized.

package.json
```diff
@@ -39,6 +39,7 @@
     ],
     "babel": "inherit"
   },
+  "types": "src/index.d.ts",
```
i8wu: Are you sure this is right? I tested this branch and needed to change it to `./src/index.d.ts` to work.

rt2zz: that seems to be the better path per http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

I will update that separate from this PR since I am still not clear how we want to move forward with the interface type + type formatting

yordis:
Could you reverse the single quote back? 

Because you modify your TS configuration your editor probably is complaining. 

This is not a TS repo so no needs for any tslint.

openscript:
Okay, so I will create a separate PR for #675 or is this issue already addressed?

yordis:
@openscript any updates?

openscript:
I've closed this pull-request because it became obsolete.

PR #682, opened 2018-01-26
Give transformers access to the full, combined reducer state

dougkeen:
For example, this is useful if you have a transformer that needs to filter items in one reducer based on values in another reducer.

Let's say we're building a chat app, where we'd like the 100 most recent messages and the sender info shown right when the app starts up (without making a trip to the server). We have a `users` reducer, which is a bunch of user data objects keyed by `userId`. We also have a `messages` reducer, which is an ordered list of message data objects, with a `senderId` property that refers to the `userId` of the sender. We'd like to build a transformer that persists only the top 100 messages in the `messages` reducer, along with all the users in the `user` reducer that correspond to the senders of those 100 messages.

Right now, this isn't possible. When persisting users in the `users` reducer, the transformer isn't able to access the `messages` reducer to figure out which users are senders of the top 100 messages.

With this PR, that transformer can be written as:
```javascript
import _ from "lodash";

const top100MessagesTransform = createTransform(
    (subState, key, fullState) => {
      if (key === "messages") {
        return _.take(subState, 100);
      } else if (key === "users") {
        const messages = _.take(fullState.messages, 100);
        const senderIdSet = _.keyBy(messages, "senderId");
        return _.pickBy(subState, (user, userId) => _.has(senderIdSet, userId));
      } else {
        return subState;
      }
    },
    (state) => state,
    {whitelist: ["messages", "users"]},
);
```

### DIFF
--- a/src/createPersistoid.js
+++ b/src/createPersistoid.js
@@ -10,9 +10,9 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
   const whitelist: ?Array<string> = config.whitelist || null
   const transforms = config.transforms || []
   const throttle = config.throttle || 0
-  const storageKey = `${config.keyPrefix !== undefined
-    ? config.keyPrefix
-    : KEY_PREFIX}${config.key}`
+  const storageKey = `${
+    config.keyPrefix !== undefined ? config.keyPrefix : KEY_PREFIX
+  }${config.key}`
   const storage = config.storage
   const serialize = config.serialize === false ? x => x : defaultSerialize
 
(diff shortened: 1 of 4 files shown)


jpshelley:
When I access the fullstate in my transform the first level is an object, so an object of my top reducers. But each reducers is a string instead of an object of key/values. Is this intended, due to my storage choice, or a bug that needs fixed? For example it looks like:
```
fullstate: Object {
   _persist: " { "version" : -1, "rehydrated": true } "
   reducerA: " { "key1": { "nestKey1": value } } "
   reducerB: " { "key2": value2 } " 
}
``` 

And after further inspection it seems like the objects just need the first/last quotation marks removed? This is happening for me on both UWP (react-native-windows) and iOS. Do not have an Android scheme.

jpshelley:
I have an issue currently with full state marked here - https://github.com/rt2zz/redux-persist/issues/871

PR #693, opened 2018-02-01
Give transformers access to the full state (i.e., across all available keys) (#682)

rt2zz:
basically https://github.com/rt2zz/redux-persist/pull/682 with fixed prettier formatting


rt2zz:
hm, nm not sure what is going on with prettier here

rt2zz:
this is now on master

Issue #690, opened 2018-02-01
Issues with flow


deecewan:
Hey,
First off, thanks a bunch for this library, and also thanks a *bunch* for adding types.

I just installed this in an existing codebase and got a whole bunch of errors, starting with:

https://gist.github.com/deecewan/1924ac53d68129148fb56cb53d1d841f

That's a whole load of flow errors from the `persistStore` function.

Edit: this could maybe be fixed by changing [this](https://github.com/rt2zz/redux-persist/blob/master/src/persistStore.js#L65) to:

```js
const persistor = {
	...createStore(...),
	flush() {},
	purge() {},
};
```

Happy to open a PR for that, if that'll fix things there. I'm running Flow v0.63 (I notice this lib is v0.59 at the moment - may be the source of issues).

Also, somewhat unrelated, but `persistReducer` makes my types go weird, because it returns `State & { _persist: PersistState }`, which doesn't match the type of my State. Is there a reason that this `_persist` item is stored in the user's state instead of the `redux-persist` redux store?

deecewan:
@rt2zz I think #697 will fix the majority of the flow errors, but still curious around why `_persist` is stored in the user's store rather than the store created with `persistStore`?

Doing this messes with my State type, because instead of being as expected, it's got an extra property on it, which I'm not too keen on adding manually because it looks like it's private.

rt2zz:
@deecewan we might change this in v6 - but its definitely not a cut and dry issue from my perspective. Here is some context:

pros:
1. keeping _persist in the reducer state makes everything atomic, we can never have state be out of sync with its persistence state (namely version). 
2. it preserves the redux concept of a "single state atom" -> everything stateful is inspectable in one place.
3. it keeps the code simple, since a store may have multiple persisted reducers, storing it along side state makes it dead simple to find the persist state for each persistoid.

cons:
1. changes state typings
2. we have to workaround combineReducers unexpected state warning
3. its unexpected for users
4. it prevents mounting multiple persists onto the same reducer (i.e. at the same mount point)

Con 4 is a big issue for me and I would definitely like to explore options for how we can resolve this. Perhaps storing _persist state in the persistor is the way to go, but then we have to worry about syncing it and persisting it.

While I am open to suggestion I think this is unlikely to change in the near future. 

In the meantime I would recommend you add the _persist state to your State type. We export the PersistState shape so you can do something like
```js
import { type PersistState } from 'redux-persist/src/types'

type State = {
  //... your state shape
[...]

Issue #685, opened 2018-01-29
redux-persist raises flow errors in latest flow version due to timer ids


jdalegonzalez:
Error: node_modules/redux-persist/lib/createPersistoid.js.flow:38
 38:       timeIterator = setInterval(processNextKey, throttle)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ IntervalID. This type is incompatible with
 23:   let timeIterator: ?number = null
                          ^^^^^^ number

Error: node_modules/redux-persist/lib/createPersistoid.js.flow:46
 46:       if (timeIterator) clearInterval(timeIterator)
                                           ^^^^^^^^^^^^ number. This type is incompatible with the expected param type of
732: declare function clearInterval(intervalId?: IntervalID): void;
                                                 ^^^^^^^^^^ IntervalID. See lib: /private/tmp/flow/flowlib_3997dbc1/core.js:732

Baransu:
I've experienced the same error. But I also have more errors than only this. I'm using flow in version 0.63.1.

Whole error message: https://gist.github.com/Baransu/8067a88c06f9d5593cadc982da06aba3

rt2zz:
look like we can resolve most of those by creating an intersection type for Persistor ie
```
type Persist = Store & { flush: Flush, purge: Purge }
```

rt2zz:
just merged https://github.com/rt2zz/redux-persist/pull/692

plz reopen if any further issues arise!

Baransu:
First fix error was fixed. Thanks. I still have a problem with
```type Persist = Store & { flush: Flush, purge: Purge }```
Here is full error:
```
Error: node_modules/redux-persist/lib/persistStore.js.flow:65
                                  v-----------
 65:   let persistor: Persistor = createStore(
 66:     persistorReducer,
 67:     undefined,
 68:     options.enhancer
 69:   )
       ^ Store. This type is incompatible with
 65:   let persistor: Persistor = createStore(
                      ^^^^^^^^^ Persistor
  Property `flush` is incompatible:
     65:   let persistor: Persistor = createStore(
                          ^^^^^^^^^ property `flush`. Property not found in
                                      v-----------
     65:   let persistor: Persistor = createStore(
     66:     persistorReducer,
     67:     undefined,
     68:     options.enhancer
     69:   )
           ^ Store

Error: node_modules/redux-persist/lib/persistStore.js.flow:65
                                  v-----------
[...]

PR #697, opened 2018-02-01
Attempt to fix Persistor creation flow error

Baransu:
Attempt to fix: #685 
Potentially fixes: #690 
Would be great to hear some feedback about this change 😃

### DIFF
--- a/src/persistStore.js
+++ b/src/persistStore.js
@@ -62,38 +62,37 @@ export default function persistStore(
     })
   }
   let boostrappedCb = cb || false
-  let persistor: Persistor = createStore(
-    persistorReducer,
-    undefined,
-    options.enhancer
-  )
 
-  persistor.purge = () => {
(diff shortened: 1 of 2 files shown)


rt2zz:
I like this change, but if I am not mistaken register and rehydrate will be broken in this case as they both depend on the store. I will try to take a look today.

Baransu:
Tests seems to work find and in my basic setup its working. I’ll be glad if you could look into it :)

rt2zz:
ah ok I had to read up on hoisting to understand how this works 👍

rt2zz:
note: I reworked it a bit to not rely on hoisting - perf should be identical but this seems like a more conservative setup:  https://github.com/rt2zz/redux-persist/commit/c090a323edaca664fce1137351e795ba4183341c

Also updated IntervalId to any for now for backwards flow compat

PR #701, opened 2018-02-03
(ops): add umd builds, libsize estimator

rt2zz:
copied from https://github.com/rt2zz/redux-persist/pull/341 - but now for v5. thanks @Andarist !

also adds a libsize estimator which will be helpful for evaluating future work. Right now we are ~4.9KB which is better than expected!

### DIFF
--- a/scripts/size-estimator.js
+++ b/scripts/size-estimator.js
@@ -1,0 +1,11 @@
+// @noflow
+const { execSync }  = require('child_process')
+const packageJson = require('../package.json')
+
+// we estimate redux size based on the content length of the minified umd build hosted by unpkg. This script is brittle but works.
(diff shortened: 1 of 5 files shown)


scripts/size-estimator.js
```diff
@@ -0,0 +1,11 @@
+// @noflow
+const { execSync }  = require('child_process')
+const packageJson = require('../package.json')
+
+// we estimate redux size based on the content length of the minified umd build hosted by unpkg. This script is brittle but works.
+let reduxSize = execSync("curl -sIL https://unpkg.com/redux/dist/redux.min.js | grep -i Content-Length | tail -1 | awk '{print $2}'").toString()
```
Andarist: you could just make a bundle with `redux` as external and just check this size locally without network requests

PR #702, opened 2018-02-03
(persistStore): add timeout option

rt2zz:
if timeout is exceeded, rehydrate will complete with an error. In the past it would simply hang indefinitely. 

timeout is default undefined -> meaning no effect.

questions:
1. this adds ~120 Bytes - is this worth it?
2. should timeout be defaulted to something large but reasonable like 5s?
3. rather than be a part of redux-persist should timeout be a feature of the storage adapter?

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "lib"
   ],
   "scripts": {
+    "ava": "BABEL_ENV=commonjs ava",
     "build": "npm run flow-copy && npm run build:commonjs && npm run build:es",
     "build:commonjs": "cross-env BABEL_ENV=commonjs babel src --out-dir lib",
     "build:es": "cross-env BABEL_ENV=es babel src --out-dir es",
--- a/src/persistStore.js
+++ b/src/persistStore.js
@@ -44,6 +44,7 @@ export default function persistStore(
   cb?: BoostrappedCb
 ): Persistor {
   let options: Object = persistorOptions || {}
+  let { timeout } = options
 
   // help catch incorrect usage of passing PersistConfig in as PersistorOptions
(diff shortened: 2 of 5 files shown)


samhunta:
Re: 

1. Worth it, would have saved me time debugging [AsyncStorage](https://github.com/facebook/react-native/issues/14101)
2. A timeout of even less (3s) would be nice for my use cases. 
3. I think it fits here well since the functionality doesn't change depending on the storage adapter

rt2zz:
👍 ok I think I agree, lets add this and default the timeout. I still want to evaluate if timeout would better live in persistReducer or here (persistStore). Will look to merge this is something comparable this week.

JulianKingman:
I like in persistReducer, since other config goes in there as well. However, I think it would be handy to pass the persist status to the persistStore callback (i.e. error or success), that way we can handle timeout cases. I like a default time, because if it's taking too long, something is likely wrong and the dev should know about it.

rimzici:
Any Updates on this?

rt2zz:
this landed via https://github.com/rt2zz/redux-persist/pull/713 - applied to persistReducer instead of persistStore 👍

rimzici:
I see, thank you.
A small confirmation, by this fix the issue of app being infinitely hung is solved, but the data is lost after the default timeout of 5000ms ?

rt2zz:
yes the data will be lost if timeout is ever reached

nico1510:
Hmm but is that really the better alternative ? I mean for my use case I would rather have the user restart the app if it doesn't render than losing all the data and continuing... Can we avoid this new logic by e.g. setting the timeout to -1 in the config or should I just pass a large number in there ?

Issue #514, opened 2017-10-28
Going backwards on API design


yordis:
> c. All constants (ex: {REHYDRATE, PURGE}) has moved to redux-persist/lib/constants instead of redux-persist/constants

I don't get why we have to put that ugly `lib` in the path, is like going backward on the package in that term.

I really don't see the point.

rt2zz:
I absolutely would like to remove that and not just for constants but all deep imports. The only question is what is the best alternative pattern? I know @Andarist has been thinking about this issue for some time.

Open to suggestions, hopefully one that does not involve maintaining a separate es5 constant file outside of src.

yordis:
@rt2zz it was working before, no?!

Andarist:
The only (known to me) possibility to have those 'cherry-picked' modules is to create proxy directories with `package.json` - like `redux-persist` [had in the past](https://github.com/rt2zz/redux-persist/blob/v4/constants/package.json).

I have wanted (and still want to) create a package to automate creation of those packages before publication or build steps. Just hadn't found enough time for this yet. I will try to sit down next weekend and create MVP for this. 

You can also ofc reexport your constants in the main entry point (`src/index.js`) so one could just do `import { constants } from 'redux-persist'`.

You could also consider using rollup to build ur package as flat bundle.

yordis:
@Andarist what would be the downside of rollup, it seems that a lot of people use it for this.

P.S: has been a long time without being involve in the JS so, what would happen if I use rollup on using your rollup version. Will rollup bundle just the piece of the app that I used? I forgot the name of that 😭

rt2zz:
cool, sounds like proxy modules is probably the way to go for now then. We might export all constants out of the main module (i.e. `import { REHYDRATE } from 'redux-persist`) and then provide proxy modules for `integration` and `storage`.

yordis:
@rt2zz is this still valid?

rt2zz:
constants are now available from the root module. Still open because we need to add proxy modules for storage and state reconcilers

Andarist:
@rt2zz could u elaborate on what proxy modules you'd like to have exactly? I might provide a PR for this.

rt2zz:
@Andarist open to suggestion, but I will list the files that I think are considered to be part of the public api:
- ./storage/index.js (although perhaps we should rename this to local.js
- ./storage/session.js
- ./stateReconciler/hardSet.js
- ./stateReconciler/autoMergeLevel1.js
[...]

PR #703, opened 2018-02-04
(api): proxy modules, some cleanup

rt2zz:


rt2zz:
I forgot we had an open discussion about this going on here: https://github.com/rt2zz/redux-persist/issues/514#issuecomment-362869310

reverted for now until we get some clarity. Going to open a more conservative version of the same thing that only splits out integration/react + readme updates

PR #704, opened 2018-02-04
Fix typo

samhunta:


rt2zz:
crap my bad forgot to checkin the new files. importing from lib will continue to work, but want to encourage using the proxy module moving forward: https://github.com/rt2zz/redux-persist/commit/f2cc14f7bfc930abb56293b2c2aec5d0fb78d358

samhunta:
👍

PR #705, opened 2018-02-04
[WIP] Flat bundles

Andarist:
This is not done, you can track progress & discuss what is being done though.


integration/react/package.json
```diff
@@ -1,7 +1,7 @@
 {
```
Andarist: as flat bundling will be a breaking change (no more reaching out to internal directory structure - which is even proposed in the main READMEs on several occasions) I'd go with simple `redux-persist/react` proxy module, I've attempted to keep both `redux-persist/integration/react` and `redux-persist/react` here but there is really no point in that

package.json
```diff
@@ -2,22 +2,24 @@
   "name": "redux-persist",
   "version": "5.6.9",
   "description": "persist and rehydrate redux stores",
-  "main": "lib/index.js",
-  "module": "es/index.js",
+  "main": "dist/redux-persist",
```
Andarist: this is matching first draft of the configs - the ones that are commented out now

package.json
```diff
@@ -2,22 +2,24 @@
...
+    "lib",
+    "dist"
   ],
+  "sideEffects": false,
   "scripts": {
     "ava": "BABEL_ENV=commonjs ava",
-    "build": "npm run flow-copy && npm run build:commonjs && npm run build:es && npm run build:umd && npm run build:umd:min",
+    "prebuild": "npm run clean",
+    "build": "rollup -c",
+    "build:old": "npm run flow-copy && npm run build:commonjs && npm run build:es && npm run build:umd && npm run build:umd:min",
```
Andarist: im not entirely sure what's proper way of supporting typings after going flat, "typings" from `flow-copy` will stop to work properly, I'd like to explore `gen-flow-files` tool (its flow builtin, but ive never used it)

TS will require some work, but at least its a sideway thing (have to be written manually anyway - so it doesnt really much, im concerned about flow because I'd like to generate them automatically and aint sure yet how to do it in a new setup)

rollup.config.js
```diff
@@ -3,41 +3,150 @@ import nodeResolve from 'rollup-plugin-node-resolve';
...
 
-const env = process.env.NODE_ENV
-const config = {
-  output: {
-    format: 'umd',
-    name: 'ReduxPersist',
-    exports: 'named',
-    sourcemap: true,  
-  },
+// const ensureArray = maybeArr => Array.isArray(maybeArr) ? maybeArr : [maybeArr];
```
Andarist: this part is very much working, although not yet supporting `storage` and `.native.js` (they can be easily added, I just havent yet got to it)

rollup.config.js
```diff
@@ -3,41 +3,150 @@ import nodeResolve from 'rollup-plugin-node-resolve';
...
+//   })
+// ]
+
+const createConfig = ({
+  input = 'src/index.js',
+  output,
+  umd = false,
+  env,
+} = {}) => ({
+  experimentalCodeSplitting: !umd,
```
Andarist: this version is interesting though, because it uses code splitting feature - this way babel helpers can be deduped between `redux-persist`, `redux-persist/storage` and `redux-persist/react` flat bundles. I've found a bug in rollup while experimenting with this though - https://github.com/rollup/rollup/issues/1943

It would also require restructuring redux-persist src directory structure slightly, because rollup maps original input filenames to output names (not including directories) and we need stable mapping between those

src/index.js
```diff
@@ -9,4 +9,8 @@ export { default as getStoredState } from './getStoredState'
 export { default as createPersistoid } from './createPersistoid'
 export { default as purgeStoredState } from './purgeStoredState'
 
+export { default as autoMergeLevel1 } from './stateReconciler/autoMergeLevel1'
```
Andarist: from what I understand it's what we have established in the other issue, correct me if im wrong

src/integration/react.js
```diff
@@ -1,5 +1,5 @@
 // @flow
-import React, { PureComponent } from 'react' // eslint-disable-line import/no-unresolved
```
Andarist: was unused

src/storage/index.js
```diff
@@ -3,3 +3,4 @@
 import createWebStorage from './createWebStorage'
 
 export default createWebStorage('local')
+export { default as session } from './session'
```
Andarist: I plan to add `#__PURE__` annotations with my plugin - https://github.com/Andarist/babel-plugin-annotate-pure-calls . So unused version should get dropped. The goal for me here was to provide a single `redux-persist/storage` entry

Andarist:
rollup has released new version with the needed bug fix. Gonna play with this PR further in following days, @rt2zz do you have any comments about what was being done so far (or about what was proposed in the comments)?

integration/react/package.json
rt2zz: I agree, `redux-persist/react` is probably best. No issues with breaking changes, although I do have a couple of breaking changes in mind, so we can stack these together for a v6

package.json
```diff
@@ -2,22 +2,24 @@
...
-  "module": "es/index.js",
+  "main": "dist/redux-persist",
+  "module": "dist/redux-persist.es",
   "types": "./src/index.d.ts",
   "repository": "rt2zz/redux-persist",
   "files": [
     "es",
     "src",
-    "lib"
+    "lib",
```
rt2zz: lib and es will go away with this correct?

package.json
rt2zz: I am not sure either. gen-flow-files may work, last I tried it was buggy but that was maybe a year ago.

Figuring this out will be a big concern however, having types is important and maintaining separate flow types is not desirable obviously (maintaining .ts declarations is already a big burden).

Is it possible to bundle w/o stripping flow types? then we could basically ship a mirrored `redux-persist.flow.js` bundle, which is basically all flow-copy-source does anyway.

rollup.config.js
rt2zz: cant say I understand. So how would src be restructured?

How would babel helpers get deduped?

src/index.js
rt2zz: 👍 I think so, but definitely want to get other opinions here. Are there any cases where we cannot rely on code splitting? Will this negatively impact any use cases?

rt2zz:
responded. 

I think the only two considerations I have moving forward are:
1. How does this look if we want to provide a umd build? The umd bundle will have a few extra bytes from the state reconcilers which I think is acceptable. Are there any other concerns?
2. How do we get flow types to work? I am pretty sure this can be done by doing a build *without* `transform-flow-strip-types`, and writing that to `redux-persist.flow.js`. 
3. because of the breaking changes, will definitely want to land this along-side other breaking changes that I am mulling over around how we store _persist, and async transforms.

package.json
Andarist: unfortunately aint sure, need to play with those things first

rollup.config.js
Andarist: At the moment with this:
```js
input: ['src/index.js', 'src/integration/react.js', 'src/storage/index.js']
```

It produces 4 files:
- index.js (what will be in main/module)
- index2.js (this should actually be `storage` for readability, that's why we need to restructure a little bit, so input file names get unique names)
- react.js 
- chunk1.js - here we have magically deduped babel helpers :)

src/index.js
Andarist: dont think so, I consider this new setup with code-splitting the most "optimized"

package.json
Andarist: actually not, i thought at first they would, but with code splitting on we do not have control over output file names so we need to output files targeting cjs and esm into 2 separate directories to avoid the file path clash

Andarist:
> How does this look if we want to provide a umd build? The umd bundle will have a few extra bytes from the state reconcilers which I think is acceptable. Are there any other concerns?

It depends what it needs to have, i suppose no react & storage in there? Or should storage be available in umd?

> How do we get flow types to work? I am pretty sure this can be done by doing a build without transform-flow-strip-types, and writing that to redux-persist.flow.js.

That's the only thing I need to experiment with - ain't sure if rollup won't chuck when parsing type nodes.

> because of the breaking changes, will definitely want to land this along-side other breaking changes that I am mulling over around how we store _persist, and async transforms.

definitely, this is a breaking change

package.json
```diff
@@ -2,30 +2,27 @@
...
   "types": "./src/index.d.ts",
   "repository": "rt2zz/redux-persist",
   "files": [
     "es",
     "src",
     "lib",
-    "dist",
     "integration",
-    "README.md"
+    "dist"
```
Andarist: README gets uploaded to npm by default

Andarist:
Ok, I got everything to work. Just one thing is not updated - typescript typings and unfortunately I won't be able to update them right now. Updating them shouldn't take much time though.

This is imho ready for final review and ready to be merged. If you have any questions regarding what I've done - just ask :)

Issue #700, opened 2018-02-02
typescript incompatible with latest version of redux-persist


ch1ll0ut1:
**Versions:**
TS: 2.7.1
redux-persist: 5.6.5

All of the sub directory imports are not possible when using typescript with redux-persist.
Example:
```javascript
import storage from 'redux-persist/lib/storage'
// error:
[ts]
Could not find a declaration file for module 'redux-persist/lib/storage'. '.../node_modules/redux-persist/lib/storage/index.js' implicitly has an 'any' type.
  Try `npm install @types/redux-persist` if it exists or add a new declaration (.d.ts) file containing `declare module 'redux-persist';`
```

rt2zz:
@ch1ll0ut1 is the issue that our types are not complete or that with deep imports TS does not know where to look for the definitions? FWIW a short term fix would be to use localforage as your storage.

ch1ll0ut1:
@rt2zz i would guess its missing typings, but cant be sure. From googling it seems it should be possible doing 
```javascript
declare module 'redux-persist/lib/storage' {
...
}
```

But on the other side, wouldnt it be better exporting the full api on the root module?

rt2zz:
got it, I think we were just missing the lib exports for these files (we had them for es), this should resolve: https://github.com/rt2zz/redux-persist/commit/4a00ed8274018d540cbd4e803216c60f94e58d92

yordis:
```
error TS4050: Return type of public static method from exported class has or is using name 'Persistor' from external module "redux-persist/es/types" but cannot be named.
```

@rt2zz I am guessing that this error is related to this,

If so, any date for publishing the fix?

yordis:
@rt2zz this one as well

```
node_modules/redux-persist/src/index.d.ts(348,46): error TS2307: Cannot find module 'react'.
```

Maybe we should add React as peerDeps? No quite sure what is the fix but I am getting this one as well

rt2zz:
@yordis I am not sure about these errors - does #706 resolve the first?

For the second it seems like maybe we need to move the type definitions for react integration out of index.d.ts and into the proxy module: https://github.com/rt2zz/redux-persist/tree/master/integration/react

Again I am not a TS user so I can only speculate, but will gladly merge fixes and cut a new release.

yordis:
> @yordis I am not sure about these errors - does #706 resolve the first?

I have `"redux-persist": "^5.6.10"`, is #706  on it?

[...]

PR #706, opened 2018-02-05
Fixed default import, removed non-existent export

rgripper:
Related to #700


i8wu:
Trying to test this out, but when installing from your branch I get `Module not found: Can't resolve 'redux-persist'`

PR #709, opened 2018-02-07
add npm-packager for releasing

yordis:
Related to #514

@rt2zz @Andarist I need your feedbacks on this.

Pretty much I enhance the npm publish and pack commands so you can specify the root of the project.

So,

What I did here was to export directly into `dist` the `commonjs` build so people could do the `require` without having to add `dist` or anything on the path.

To be honest,

I am not quite sure which one is the build that have to be flatted out (that is why I need you to double check this) but the one that we need to flat out, we can just put the output directly into `dist` folder and done.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -272,6 +272,50 @@
         }
       }
     },
+    "@straw-hat/npm-packager": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@straw-hat/npm-packager/-/npm-packager-1.0.0.tgz",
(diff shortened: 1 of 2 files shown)


yordis:
Yeah this one, you will need to give up to one of them 🤔🤔🤔

Andarist:
Not sure what do you mean, could you rephrase?

yordis:
@Andarist the `commonjs` one do not need to have the `dist` on the path because it is flatted out.

Sorry for the inconvenience, I thought this may help.

PR #713, opened 2018-02-10
(persistReducer): timeout

rt2zz:
alternative to https://github.com/rt2zz/redux-persist/pull/702/files but with default 5000ms timeout and applied to persistReducer instead of persistStore


rt2zz:
note: this adds about 200 bytes to the lib

PR #718, opened 2018-02-12
feature: make PersistGate take function as a child component

### LINKED ISSUE #655: PersistGate: animate loading component
Hi, is there a way I can animate loading component, I know I can start an animation on life cycle events such as componentDidMount, but is there a way I can set up any other transition? or wait for it to be finished before lifting the gate?

Thanks!

wachunei:
Resolves #655 

PersistGate now takes a function as a child component:
```jsx 
<Provider store={store}>
  <PersistGate persistor={persistor}>
    {bootstraped => (
      // custom component 
    )}
  </PersistGate>
</Provider>
```

In case both `loading` prop and children as function are passed, `loading` prop will be ignored.

### DIFF
--- a/src/integration/react.js
+++ b/src/integration/react.js
@@ -5,7 +5,7 @@ import type { Persistor } from '../types'
 
 type Props = {
   onBeforeLift?: Function,
-  children?: Node,
+  children?: Node | Function,
   loading?: Node,
   persistor: Persistor,
 }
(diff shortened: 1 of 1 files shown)


rt2zz:
tests would be awesome

I will add some basic docs, and also a dev only validation

No need to generate build files, that is done automatically upon release.

Thanks!

xstable:
Would be nice, if you could bundle a release: @rt2zz

PR #723, opened 2018-02-15
fix: use self when referencing the global scope

zacharygolba:
Currently you are not able to use `redux-persist` inside of a [Worker](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). This is due to `window` not being defined in [`WorkerGlobalScope`](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope). This PR fixes the compatibility problem by using [`self`](https://developer.mozilla.org/en-US/docs/Web/API/Window/self) instead of `window`.

### DIFF
--- a/src/integration/getStoredStateMigrateV4.js
+++ b/src/integration/getStoredStateMigrateV4.js
@@ -24,12 +24,12 @@ export default function getStoredState(v4Config: V4Config) {
 const KEY_PREFIX = 'reduxPersist:'
 
 function hasLocalStorage() {
-  if (typeof window !== 'object' || !('localStorage' in window)) {
+  if (typeof self !== 'object' || !('localStorage' in self)) {
     return false
   }
 
   try {
-    let storage = window.localStorage
+    let storage = self.localStorage
     const testKey = `redux-persist localStorage test`
     storage.setItem(testKey, 'test')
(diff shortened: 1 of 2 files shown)


rt2zz:
nice, thank you for this.

Are there any possible complications with using self? browser compatibility issues?

zacharygolba:
No problem! Great work with this library. 😃 

From the looks of this [compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Window/self#Browser_compatibility) `self` is supported in all major browsers (including IE11).

If a case comes up where compatibility is an issue for a user, they could naively polyfill `self` with something along these lines:

```javascript
(function (ctx) {
  if (!('self' in ctx)) {
    ctx.self = ctx;
  }
})(this);
```

rt2zz:
👍 

I do consider pulling storage out of redux-persist since it is browser specific and comes with some additional overhead that not all users need (e.g. private browsing fallback). 

Generally I recommend using localForage (which plugs in without any modification) but that is a non trivially sized package.

Anyway thanks again!

zacharygolba:
That makes sense. Happy to help!

rt2zz:
just released in v5.9.0

PR #750, opened 2018-03-07
Add persist() and pause() to Typescript definition

raymondsze:
persist and pause function is missing in Persistor in index.d.ts.


rt2zz:
thanks!

PR #754, opened 2018-03-09
Use flow-typed for better typing

kogai:
## Why

It doesn't might be caused by `redux-persist`, but although type system of `flow`, especially a collector of type-check targets, there exists potential type errors.

Ideally, `flow` should report an error for the absence of a type definition of `redux`, but not.
I'd say it is an unexpected behavior of `flow`, but it would be better to prevent such problem of the type in the library, especially for users using the library such useful.

And then, I attempt to use `flow-typed` which providing actual type definitions from the ecosystem.
I think it would be better, but if you (or actual users) don't think so, feel free to point out and I'll withdraw my PR.


kogai:
Due to it does not seems to be reviewed, I'll close the PR.

PR #760, opened 2018-03-13
fixes #646

### LINKED ISSUE #646: add react to peerDependencies
Right now, using JSPM, a forced map is needed to resolve `import React, { PureComponent } 'react'` because it doesn't list react as a dependency (or peer).

yordis:


yordis:
cc: @rt2zz this should fix the issue with the types, hopefully.

yordis:
@rt2zz Could we get this in?

PR #766, opened 2018-03-16
How to use custom migration to keep our application stable for offline purposes

lusan:
Although the docs are good, but I felt the README needed an example of a live use case on how we can use the library to perform custom migrations. 
The goal is to help users understand the need for migration and how they can use this feature to keep their application stable for offline purposes. Misuse can easily lead to bad user experience and cause unknown issues in production.

Hope this small addition helps everyone use this amazing library.


lusan:
@rt2zz can you please review it

Issue #753, opened 2018-03-08
Argument of type 'Reducer<{}>' is not assignable to parameter of type 'BaseReducer<{}, AnyAction>'


Icehunter:
```
ERROR in [at-loader] ./ClientApp/application.tsx:38:4 
    TS2345: Argument of type 'Reducer<{}>' is not assignable to parameter of type 'BaseReducer<{}, AnyAction>'.
  Types of parameters 'state' and 'state' are incompatible.
    Type 'void | {}' is not assignable to type '{}'.
      Type 'void' is not assignable to type '{}'.
```

Relevant Code:

store.ts
```javascript
import { combineReducers } from 'redux';

import * as User from './User';

export interface ApplicationState {
  user: User.UserState;
}

export const reducers = combineReducers({
  user: User.reducer
});

```

application.tsx
```javascript
import { reducers } from './store';

const reducer = combineReducers(Object.assign({}, {
  store: reducers
}, {
  routing: routerReducer
}));

const persistedReducer = persistReducer({
  key: 'root',
  storage: storage,
}, reducer);

const store = compose(applyMiddleware(
  middleware,
  thunk,
  logger
))(createStore)(persistedReducer);

const persistedStore = persistStore(store);
```

My User reducer is exported as follows:
```
export const reducer: Reducer<UserState> = (state: UserState, action: any) => {}
```

Using 5.5.0 with typescript 2.6.2 produces no error; but if I update to 5.9.1; I get this one.

dawnmist:
The same issue occurs with the definition of persistReducer. The reducer function returned is missing the `| undefined` for the state, so redux 4.0 rejects the persistedReducer when you try to create the store from it.

Since you already have a pull request for updating this bit, do you want to add the fix for the persistReducer function too?

```javascript
    export function persistReducer<S, A>(config: PersistConfig, baseReducer: BaseReducer<S, A>): (s: S | undefined, a: A) => S & PersistPartial;
```

PR #778, opened 2018-03-26
Fix BaseReducer typescript definition to be redux-compatible

doomsower:
Current typescript definitions are not compatible with current redux definitions: https://github.com/rt2zz/redux-persist/issues/753

This is how reducer definition looks like in [redux 3.7.2](https://github.com/reactjs/redux/blob/8f60ba321e8ba5fa71d60fa35573c2cdf9c0d852/index.d.ts#L57)

```ts
export type Reducer<S> = (state: S, action: AnyAction) => S;
```

In redux 4.0.0-beta the definition will be changed to one that involves undefined (which is still incompatible with `void` ([link](https://github.com/reactjs/redux/blob/9f250e0828083c15d0b9dd3d283d7f3f9223246c/index.d.ts#L59)):
```ts
export type Reducer<S = any, A extends Action = AnyAction> = (state: S | undefined, action: A) => S;
``` 

I think it is better to stick to current stable version of redux.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -342,7 +342,7 @@ declare module "redux-persist/es/persistReducer" {
     import { PersistState, PersistConfig } from "redux-persist/es/types";
     // persistReducer
     export interface PersistPartial { _persist: PersistState }
-    export type BaseReducer<S, A> = (state: S | void, action: A) => S;
+    export type BaseReducer<S, A> = (state: S, action: A) => S;
     /**
      * It provides a way of combining the reducers, replacing redux's @see combineReducers
      * @param config persistence configuration


src/index.d.ts
```diff
@@ -342,7 +342,7 @@ declare module "redux-persist/es/persistReducer" {
     import { PersistState, PersistConfig } from "redux-persist/es/types";
     // persistReducer
     export interface PersistPartial { _persist: PersistState }
-    export type BaseReducer<S, A> = (state: S | void, action: A) => S;
+    export type BaseReducer<S, A> = (state: S, action: A) => S;
```
dawnmist: I think this should probably be `(state: S | undefined, action: A) => S` in order to match up with the redux 4.0 types. After all, the reducer is initialized by passing `undefined` into the reducer function.

osdiab:
Anything holding this up?

Issue #634, opened 2017-12-20
Migrations are triggered only for integer version numbers


bugron:
Migrations happen only for integer version numbers. For example, migrations below will be applied:
```js
const migrations = {
  0: state => {
    //...
  },
  1: state => {
    //...
  },
  2: state => {
    //...
  },
  // ect.
}
```
Though, those migrations will not:
```js
const migrations = {
  '0.0.1': state => {
    //...
  },
  '0.0.2': state => {
    //...
  },
  // ect.
}
```
I want to tie `redux-persist`'s config version to my app's version and in this case for not a major version change no migrations will be triggered. For instance, a migration from `0.0.1` (`inboundVersion`) to `0.0.2` (`currentVersion`) won't happen because, as per [this line](https://github.com/rt2zz/redux-persist/blob/master/src/createMigrate.js#L38) in `createMigrate.js` the `key` (in this example, `0.0.2`) will be parsed to `0` and then `filter` will successfully filter out the `0.0.2` key. This will result to no migration applied.
If I remove that line from `createMigrate.js`, then I'm able to apply migrations even from `0.0.1-alpha` to `0.0.1-beta` thanks to JS string comparison. Or, maybe, `semver` package (or similar) can be used to compare versions to not to rely on a simple string comparison.
The thing is that I can still migrate my app from `0.0.1` to `0.0.2` by assigning config version `0` to the first one and `1` to the second one. It just seems more reasonable to me to tie app's version to config's version.
What do you think? It there anything you can recommend? Is this something you can consider to change?

PR #805, opened 2018-04-23
Allow semver version comparison

nickofthyme:
Currently using `parseInt` prior to comparing versions to determine if migration is needed. Allow for comparison between semver string versions to trigger migrations.

Addresses issue #634


nickofthyme:
Close to use major versions for all migrations.

Issue #811, opened 2018-04-27
State is not updated on storage under these conditions


Dante-101:
I figured out two conditions which were giving rise to unexpected persistence issues.

1. If the last transformer returns undefined for the last modified key, the writing is completely skipped. It is an issue if no more updates are sent to the state. Even flush does not persist the state.
This happens in `createPersistoid.js`, line 49-53.
```
var key = keysToProcess.shift();
var endState = transforms.reduce(function (subState, transformer) {
    return transformer.in(subState, key, lastState);
}, lastState[key]);
if (typeof endState !== 'undefined') stagedWrite(key, endState);
```
I created a pull request which fixes this behavior:
https://github.com/rt2zz/redux-persist/pull/812 .

2. On state update, if I end up deleting a key, the update function doesn't take that into account and the new change is not persisted. Some comments suggest setting it to null but I don't want to assign a null to a deleted key. I have 100s of keys that keep changing for each user on a day-to-day basis and I have to delete it else the state will keep getting bigger.

I created another pull request which fixes both these behaviour.
https://github.com/rt2zz/redux-persist/pull/813

I created a separate pull request because I saw this thread https://github.com/rt2zz/redux-persist/issues/492 and was not confident how it would play out with dynamic modification of reducers.

Dante-101:
I needed the fix urgently so I created a branch with the compiled distributable dirs. Get the fix by:

`npm install github:Dante-101/redux-persist#stage-write-fix-dist`

erkrapide:
I would like to see @Dante-101's fix in a new version release. Is that possible?

rt2zz:
resolved in https://github.com/rt2zz/redux-persist/pull/812

PR #812, opened 2018-04-27
fixed an edge case for staged write

Dante-101:
Fixes this issue - https://github.com/rt2zz/redux-persist/issues/811


rt2zz:
thanks! good catch

Issue #831, opened 2018-05-28
Support downgrading


vincenzovitale:
It would be great to have support for downgrade.
So changing this line:
https://github.com/rt2zz/redux-persist/blob/master/src/createMigrate.js#L33

into something that take a downgrade array of functions and apply a reverse reduce function.

arnaudambro:
I agree, especially when you create branches with new features in the app impacting the state, and that you want to dump afterwards.

PR #832, opened 2018-05-28
Support for downgrade migrations

vincenzovitale:
Added support for downgrade migrations with an optional argument

```js
downgradeMigrations?: MigrationManifest
```

in createMigrate.

Should cover:
https://github.com/rt2zz/redux-persist/issues/831

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -21,11 +21,31 @@ const migrations = {
   }
 }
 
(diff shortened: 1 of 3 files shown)


casvil:
is this the support for downgrade migrations? why is not closed?

cbh6:
:exclamation:

PR #834, opened 2018-05-30
Typescript BaseReducer type definition

osdiab:
https://github.com/rt2zz/redux-persist/pull/778 is stale so I'm taking it into my hands.

First commit just makes BaseReducer type work with Redux types.

Second commit goes further to remove it altogether, but if we want to keep `BaseReducer` for backwards compatibility I can revert that.

I don't think there's any problem with directly referring to Redux types, but if so I can change it to primitive types if you want.

### DIFF
--- a/src/index.d.ts
+++ b/src/index.d.ts
@@ -343,13 +343,12 @@ declare module "redux-persist/es/persistReducer" {
     import { Reducer, Action } from "redux";
     // persistReducer
     export interface PersistPartial { _persist: PersistState }
-    export type BaseReducer<S, A extends Action> = Reducer<S, A>;
     /**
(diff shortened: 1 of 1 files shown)


osdiab:
@dawnmist requesting you since you were on the previous pull.

osdiab:
Closes https://github.com/rt2zz/redux-persist/issues/753.

osdiab:
Tested works in my code.

osdiab:
Cool, what’s the process for getting things merged and deployed? I’d love to make use of this code.

osdiab:
@rt2zz

rt2zz:
thanks @osdiab ! I have been away but getting back to things now.

PR #849, opened 2018-06-23
persistReducer is defined but never used.

joshlacey:
It looks like you're using the persistorReducer here instead of the persistReducer that you imported.

### DIFF
--- a/src/persistStore.js
+++ b/src/persistStore.js
@@ -11,7 +11,7 @@ import type {
 } from './types'
 
 import { createStore } from 'redux'
-import persistReducer from './persistReducer'
+// import persistReducer from './persistReducer'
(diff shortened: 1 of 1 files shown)


src/persistStore.js
```diff
@@ -11,7 +11,7 @@ import type {
 } from './types'
 
 import { createStore } from 'redux'
-import persistReducer from './persistReducer'
+// import persistReducer from './persistReducer'
```
robinpokorny: If it is never used, I'd just remove it.

rt2zz:
removed thanks! https://github.com/rt2zz/redux-persist/commit/f6f6d642fb3e3f32544189a336442726344f323a

Issue #876, opened 2018-07-29
Does not render from the server for isomorphic applications


codejamninja:
Why are we subscribing to the persistor after the component renders? This breaks redux-persist for isomorphic applications because the server does not rerender since the html has already been sent to the client.

```js
componentDidMount() {
  this._unsubscribe = this.props.persistor.subscribe(
    this.handlePersistorState
  )
  this.handlePersistorState()
}
```
Subscribing to the persistor before the first (and only) render enables proper isomorphic support.

```js
componentWillMount() {
  this._unsubscribe = this.props.persistor.subscribe(
    this.handlePersistorState
  )
  this.handlePersistorState()
}
```

https://github.com/rt2zz/redux-persist/blob/master/src/integration/react.js#L27-L32

codejamninja:
If you want this fix immediately, you can add the following to the `dependencies` section of your `package.json` file.

```js
"redux-persist": "git+https://github.com/codejamninja/redux-persist.git#npm",
```

PR #877, opened 2018-07-29
Fix server isomorphic render

codejamninja:
Subscribe to persistor before initial render to enable isomorphic support

Fixes issue #876

### DIFF
--- a/src/integration/react.js
+++ b/src/integration/react.js
@@ -24,7 +24,7 @@ export class PersistGate extends PureComponent<Props, State> {
   }
   _unsubscribe: ?Function
 
-  componentDidMount() {
+  componentWillMount() {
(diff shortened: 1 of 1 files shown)


codejamninja:
If you want this fix immediately, you can add the following to the `dependencies` section of your `package.json` file.

```js
"redux-persist": "git+https://github.com/codejamninja/redux-persist.git#npm"
```

src/integration/react.js
```diff
@@ -24,7 +24,7 @@ export class PersistGate extends PureComponent<Props, State> {
   }
   _unsubscribe: ?Function
 
-  componentDidMount() {
+  componentWillMount() {
```
rt2zz: componentWillMount is being deprecated I think? is there another way to achieve the same effect?

PR #887, opened 2018-08-15
Improve Storage Engines section in README

rodrigobdz:
Add missing information regarding dynamic storage engine selection when using redux-persist/lib/storage.
I added the info that was already available in [code example](https://github.com/rt2zz/redux-persist/blame/208c6b112ead87b3701dfacaae2cdbe78377775a/README.md#L24).

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,7 +212,7 @@ The createTransform function takes three parameters.
 3. Config.
 
 ## Storage Engines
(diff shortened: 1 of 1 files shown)


rt2zz:
we are going to remove the default AsyncStorage as it has moved out of react-native. Thanks!

PR #903, opened 2018-09-06
(docs) Add web extension storage engine link

ssorallen:
[redux-persist-webextension-storage][0] is a redux-persist storage
engine for [WebExtension storage][1]. Both Chrome and Firefox implement
the same API, so the engine can be used in both browsers.

I created redux-persist-webextension-storage to use inside [Tab Wrangler][2], and it's currently used in the production version to wrap Chrome/Firefox web extension local storage.

[0]: https://github.com/ssorallen/redux-persist-webextension-storage
[1]: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage
[2]: https://github.com/tabwrangler/tabwrangler


rt2zz:
awesome! haha tab wrangler looks cool. 

This is what I am working with atm
screen shot 2018-12-09 at 3 00 02 pm

ssorallen:
There are a bunch of tab management extensions out there, but the Tab Wrangler model works best for me: let it automatically close everything. I realized I don't need most of those tabs!

PR #911, opened 2018-10-04
(persistoid): Do no unnecessary writes with undefined values

myrjola:
Earlier, unchanged undefined values caused unnecessary writes.

These changes also introduce unit tests for `Persistoid.update`.

### DIFF
--- a/src/createPersistoid.js
+++ b/src/createPersistoid.js
@@ -37,7 +37,7 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
     //if any key is missing in the new state which was present in the lastState,
     //add it for processing too
     Object.keys(lastState).forEach(key => {
-      if (state[key] === undefined) {
+      if (lastState[key] !== undefined && state[key] === undefined) {
         keysToProcess.push(key)
       }
     })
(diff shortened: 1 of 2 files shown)


src/createPersistoid.js
```diff
@@ -37,7 +37,7 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
     //if any key is missing in the new state which was present in the lastState,
     //add it for processing too
     Object.keys(lastState).forEach(key => {
-      if (state[key] === undefined) {
+      if (lastState[key] !== undefined && state[key] === undefined) {
```
rt2zz: this check makes sense. Can you double check how this will interact with the recently merged: https://github.com/rt2zz/redux-persist/pull/928/files

Your tests look good, and run and pass. So my guess is we can simply add `lastState[key] !== undefined` as one more item in the conditional.

rt2zz:
@myrjola 	this looks great. I just resolve the merge conflict. Can you double check everything still checks out (see my inline comment)? The logic makes sense and I am planning to merge this soon.

myrjola:
Looks good!

PR #912, opened 2018-10-04
Misleading error message when initializing redux-persist server side.

PepijnSenders:
Memory storage is usually considered as something different than actually not storing it at all. I feel like renaming it to noop storage fits better.


PepijnSenders:
@rt2zz wdyt?

rt2zz:
thanks!

Issue #913, opened 2018-10-05
Nested persisted reducers does not work with Immutable


essensx:
There is an issue with nested persistant reducers with Immutable due to object being frozen.

`Uncaught (in promise) TypeError: Cannot add property _persist, object is not extensible`

```
if (newState === restState) return state;else {
      newState._persist = _persist;
      return conditionalUpdate(newState);
    }
```

Any ideas on how to address this?

prescottprue:
I am also seeing this without using immutable

mariogillazaro:
I get `Exception: TypeError: 'caller', 'callee', and 'arguments' properties may not be accessed on strict mode functions or the arguments objects for calls to them at Function.remoteFunction (<anonymous>:2:14)` inside the fields `args` and `caller` of the `register` inside the `persist/PERSIST` action

Unforgiven-wanda:
@mariogillazaro I find myself in the same situation. Mind explaining how you fixed it?
Thank you

mustafamagdy:
Any idea how to fix this issue?
I even remove `ImmutableJS`, and I got this error 
`Attempted to assign to readonly property.`

mustafamagdy:
Will this be available in npm installable version or just from this repo?

aguynamedben:
Waiting on new release, tracking new release in #980.

mustafamagdy:
Is there a roadmap or a schedule when the new release will be pushed to npm, even in alpha, because of this still a big issue in dealing with the state?

PR #915, opened 2018-10-10
(persistReducer) Fix frozen _persist bug

essensx:
Related with problems when using immutable or any other state manager that freezes the object which includes `_persist` object. This pull request addresses the issue.

Related: https://github.com/rt2zz/redux-persist/issues/913


sanjaypojo:
@Anahkiasen can you elaborate on how you're getting it to work with `immer`? I'm facing the same issue.

mustafamagdy:
@JustinLivi  I think the PR is cleaner, so instead of spreading the object in each reducer it is fixed in the persisted state update

PR #917, opened 2018-10-15
Modified read me

linux08:
modified read me to show an example of how to import persistor and store

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,11 +32,11 @@ const persistConfig = {
 
 const persistedReducer = persistReducer(persistConfig, rootReducer)
 
-export default () => {
-  let store = createStore(persistedReducer)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,6 +45,7 @@ If you are using react, wrap your root component with [PersistGate](./docs/Persi
 import { PersistGate } from 'redux-persist/integration/react'
 
 // ... normal setup, create store and persistor, import components etc.
+//example : import { store, persistor } from './configureStore'
```
aguynamedben: I think this would be clear if you just move `import { store, persistor } from './configureStore'` to right under the PersistGate import above. If you do that I think we can remove line 47, then I'll merge it.

linux08: Just did that, kindly check.

PR #928, opened 2018-11-06
Stop process blacklisted and already pushed keys from lastState

zarv1k:
Fixes the issue with processing blacklisted keys in redux-persist@5.10.0 caused by #813.


rt2zz:
thank you!

PR #932, opened 2018-11-26
Unify createTransform and Transform types

### LINKED ISSUE #842: [flow] createTransform is incompatible with type Transform
The `Transform` type has following interface:
```
export type Transform = {
  in: (Object | string, string, Object | string) => Object,
  out: (Object | string, string, Object | string) => Object,
  config?: PersistConfig,
}
```
But `createTransform` accepts only objects for both `state` and `fullState`:
```
export default function createTransform(
 .....
) {
  .....
  return {
    in: (state: Object, key: string, fullState: Object) =>
      .....
    out: (state: Object, key: string, fullState: Object) =>
      .....
  }
}
```
Is there any reason why `Transform` type accepts both `string` and `Object` ? Maybe `Transform` interface should be reduced to what `createTransform` offers?

lao9:
Fixes #842


aguynamedben:
I'm not well-versed in this codebase, but could you explain why this PR is so massive? There are tons of new, large files added. Did you reorganize the project?

lao9:
Thanks for pointing that out @aguynamedben - I meant to change one file in this particular PR! 

I don't think that this repo is maintained anymore, so to ensure flow was passing on my current project I ran through the build steps on my forked repo so I could use my fork on my own project. Just going to close this PR I think for now.

aguynamedben:
@lao9 Ah okay cool, thanks haha. Yeah we use redux-persist and I noticed this project was seems abandoned, and was browsing some PRs to size up how much work it'd be to help maintain it in some meaningful way.

PR #949, opened 2018-12-14
Add configurable handler to onWriteFail

ninjacarr:
When the storage engine fails to persist data, the exception from setItem() is by default silently ignored. This can happen when storage is full, for example when a browser refuses to allocate enough storage due to the quota being exceeded. (In Chrome, the quota is variable based on a percentage of current free disk space.)

This PR adds an optional function to PersistConfig that will be notified when this happens, which enables the application to deal with it or notify the user.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -63,8 +63,12 @@ The Persistor is a redux store unto itself, plus
   debug?: boolean, // true -> verbose logs
   stateReconciler?: false | StateReconciler, // false -> do not automatically reconcile state
   serialize?: boolean, // false -> do not call JSON.parse & stringify when setting & getting from storage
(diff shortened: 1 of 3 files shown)


ninjacarr:
For anyone searching error messages: this provides a way to deal with the **Error storing data** message seen in console.

schallerala:
Did you try it out by disabling the local storage in Google Chrome for example? (Going into its settings and disabling cookies (therefore local storage) for that particular site, what will throw an error when trying to access ```window.localStorage```)
Because, when I tried using ```writeFailHandler```, it didn't do anything.

The reason is because the ```getStorage``` returns a ```noopStorage``` when ```hasStorage``` fails. ```noopStorage``` that will never fail when used inside the Promise.

PR #958, opened 2019-01-03
add customizing serializer/deserializer config

scott1028:
add serializer/deserializer config instead of default serializer/deserializer, such as support `serializer-javascript`

### DIFF
--- a/src/createPersistoid.js
+++ b/src/createPersistoid.js
@@ -16,7 +16,10 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
     config.keyPrefix !== undefined ? config.keyPrefix : KEY_PREFIX
   }${config.key}`
   const storage = config.storage
-  const serialize = config.serialize === false ? x => x : defaultSerialize
+  let serialize = config.serialize === false ? x => x : defaultSerialize
+  if (typeof config.serialize === 'function') {
+    serialize = config.serialize
+  }
   const writeFailHandler = config.writeFailHandler || null
 
   // initialize stateful values
--- a/src/createPersistoid.js
+++ b/src/createPersistoid.js
@@ -16,9 +16,13 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
     config.keyPrefix !== undefined ? config.keyPrefix : KEY_PREFIX
   }${config.key}`
   const storage = config.storage
-  let serialize = config.serialize === false ? x => x : defaultSerialize
-  if (typeof config.serialize === 'function') {
+  let serialize
+  if (!config.serialize) {
+    serialize = defaultSerialize
+  } else if (typeof config.serialize === 'function') {
     serialize = config.serialize
+  } else {
+    serialize = x => x
   }
   const writeFailHandler = config.writeFailHandler || null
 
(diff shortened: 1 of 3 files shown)


scott1028:
Sometime I want to persist function in store of redux even if I reload the browser, so I modified the config object to accept customizing serializer/deserializer, such as using [serialize-javascript](https://github.com/yahoo/serialize-javascript)

src/createPersistoid.js
```diff
@@ -16,7 +16,10 @@ export default function createPersistoid(config: PersistConfig): Persistoid {
     config.keyPrefix !== undefined ? config.keyPrefix : KEY_PREFIX
   }${config.key}`
   const storage = config.storage
-  const serialize = config.serialize === false ? x => x : defaultSerialize
+  let serialize = config.serialize === false ? x => x : defaultSerialize
+  if (typeof config.serialize === 'function') {
+    serialize = config.serialize
+  }
```
aguynamedben: Should this be written a little bit differently so it's not deciding a value then immediately overwriting it? Something like:

```js
let serialize;
if (config.serialize === true) {
  serialize = defaultSerialize;
} else if (config.serialize === false) {
  serialize = x => x;
} else if (typeof config.serialize === 'function') {
  serialize = config.serialize;
} else {
  throw Error('config.serialize must be true, false, or a function');
}
```

src/getStoredState.js
```diff
@@ -13,7 +13,10 @@ export default function getStoredState(
   }${config.key}`
   const storage = config.storage
   const debug = config.debug
-  const deserialize = config.serialize === false ? x => x : defaultDeserialize
+  let deserialize = config.serialize === false ? x => x : defaultDeserialize
+  if (typeof config.deserialize === 'function') {
+    deserialize = config.deserialize
+  }
```
aguynamedben: (same here re: comment above)

src/types.js
```diff
@@ -23,7 +23,8 @@ export type PersistConfig = {
   stateReconciler?: false | Function,
   getStoredState?: PersistConfig => Promise<PersistedState>, // used for migrations
   debug?: boolean,
-  serialize?: boolean,
+  serialize?: boolean | Function,
+  deserialize?: Function | null,
```
aguynamedben: Shouldn't `deserialize?` and `serialize?` have the same value here?

src/createPersistoid.js
aguynamedben: I guess there is no `config.serialize === true`, people can just pass `false` or `function`, so maybe...

```js
let serialize;
if (config.serialize === false) {
  serialize = x => x;
} else if (typeof config.serialize === 'function') {
  serialize = config.serialize;
} else {
  serialize = defaultSerialize;
}
```

scott1028: Good Idea! This make sense more. At first I just want to modify a little code to implement the feature I want.
Is this performace better, right?

src/types.js
scott1028: OK, It should be exactly same.

src/createPersistoid.js
aguynamedben: I suggested that for readability, not performance. Overwriting `serialize` after it's set by some other logic seems like it could lead to misunderstandings about what's going on with `serialize` (i.e. someone reading the code might not realize we're going to change the value after setting it).

Thanks for the contribution!

PR #962, opened 2019-01-08
Fix HMR example

mattlubner:
The existing HRM example contains an error, in that webpack's `module.hot.accept()` function doesn't invoke the passed callback on success when the dependencies argument is omitted (it's only invoked on errors).

See: https://webpack.js.org/api/hot-module-replacement/#accept-self-

I've also changed the import line for `nextRootReducer` to explicitly reference the default export, since this file is written in ES modules syntax, and therefore it stands to reason that './path/to/reducer' is also using ES modules syntax.


aguynamedben:
@mattlubner Thank you for the contribution. Merging now. 🙏

Issue #953, opened 2018-12-20
Flow errors 


kareemsalah227:
Reopening this again, the last issues on flow issues is from March 2018, there's a lot of Flow errors.

Also when integrating with React and using <PersistGate>, this specific error is thrown:
```
Error ┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈ src/integration/react.js:17:8

Cannot extend PureComponent [1] with PersistGate because:
 • in the return value of property render:
    • Either undefined [2] is incompatible with null [3].
    • Or property @@iterator is missing in undefined [2] but exists in $Iterable [4].
 • in the return value of property render:
    • Either undefined [5] is incompatible with null [3].
    • Or property @@iterator is missing in undefined [5] but exists in $Iterable [4].

     src/integration/react.js
 [2]  8│   children?: Node | Function,
 [5]  9│   loading?: Node,
       :
     14│   bootstrapped: boolean,
     15│ }
     16│
 [1] 17│ export class PersistGate extends PureComponent<Props, State> {
     18│   static defaultProps = {
     19│     loading: null,
     20│   }
     21│
     22│   state = {
     23│     bootstrapped: false,
     24│   }
     25│   _unsubscribe: ?Function
     26│
     27│   componentDidMount() {
     28│     this._unsubscribe = this.props.persistor.subscribe(
     29│       this.handlePersistorState
     30│     )
     31│     this.handlePersistorState()
       :
     52│
     53│   render() {
     54│     if (process.env.NODE_ENV !== 'production') {
     55│       if (typeof this.props.children === 'function' && this.props.loading)
     56│         console.error(
     57│           'redux-persist: PersistGate expects either a function child or loading prop, but not both. The loading prop will be ignored.'
     58│         )
     59│     }
     60│     if (typeof this.props.children === 'function') {
     61│       return this.props.children(this.state.bootstrapped)
     62│     }
     63│
     64│     return this.state.bootstrapped ? this.props.children : this.props.loading
     65│   }
     66│ }
     67│

     /tmp/flow/flowlib_fdae0d/react.js
 [3] 14│   | null
       :
[...]

PR #974, opened 2019-01-23
Add correct PersistGate defaultProps to support newer flow versions 

erksch:
As mentioned in #953 there are type errors with some flow versions in the PersistGate component.

The problem with flow is described here:
https://github.com/facebook/flow/issues/1660

A possible solution is described [later in the issue](https://github.com/facebook/flow/issues/1660#issuecomment-386619834) and was already proposed by the creator of issue #953.

So what I did is adding `children: null` to the defaultProps and making `loading and children` not optional anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
+- [PR #974](https://github.com/rt2zz/redux-persist/pull/974) **Add correct PersistGate defaultProps to support newer flow versions.** Adds `children` to defaultProps in `PersistGate` component and removes the optional (`?`) flag in order to be compatible with flow type checking (more explanation can be found in [this flow issue](facebook/flow#1660)). Fixes [#953](https://github.com/rt2zz/redux-persist/issues/953)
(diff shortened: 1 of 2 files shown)


kareemsalah227:
I only have a quick comment, correct me if I'm wrong please.

You can provide a loading component or you can ignore it, so it's kind of "optional", I think this might be the reason it was set optional in the flow type? If that is the case, is there another solution to make it optional and get rid of the flow error? (maybe by providing a default value or something).

Please let me know what do you think.

erksch:
Removing the `?` is okay here because flow will mark them as optional automatically when defaultProps are defined. You can read about it here: https://github.com/facebook/flow/issues/1660#issuecomment-386619834

erksch:
And there seems to be no other option than removing the optional flag :/

kareemsalah227:
Actually there is, I'll create a quick PR for that...

kareemsalah227:
Ops, looks like I don't have permissions to push a branch to the repo, what should I do? @erksch

kareemsalah227:
Nevermind, I've never contributed to an open source repo before, pardon my ignorance.
I forked the repo and created a cross-repo PR.
https://github.com/rt2zz/redux-persist/pull/975

There's a decent description in there, let me know what do you think @erksch .

PR #975, opened 2019-01-23
Checks children and loading props

kareemsalah227:
### What does this PR do?
  Fixes the flow type issue with optional `children` and `loading`.

### Related conversation:
- This is the issue #953 
- This PR has some conversation comments as well https://github.com/rt2zz/redux-persist/pull/974

### Description:
Before the PR changes, the return line was causing problems, since both props are optional `children` and `loading`, doing `return condition ? this.props.children : this.props.loading` is causing Flow to complain, the idea is to set those properties a default value, since `this.props` can't actually be set in the component code, a separate variables should be created that can be set the default values.

The same behavior is done but returning null instead of setting the default value which is null.

### Important comment:
Setting both properties to required and setting default values for them is actually not a semantic solution, documentation is going to show that they are required while they aren't, thus they have to stay optional and the initialization has to be done explicitly somewhere else, because it doesn't make sense to initialize this.props at least to me, should Facebook address that in their Flow implementation?

### DIFF
--- a/src/integration/react.js
+++ b/src/integration/react.js
@@ -51,16 +51,27 @@ export class PersistGate extends PureComponent<Props, State> {
   }
 
   render() {
+    const { children, loading } = this.props
+
     if (process.env.NODE_ENV !== 'production') {
-      if (typeof this.props.children === 'function' && this.props.loading)
+      if (typeof children === 'function' && loading)
         console.error(
           'redux-persist: PersistGate expects either a function child or loading prop, but not both. The loading prop will be ignored.'
         )
     }
-    if (typeof this.props.children === 'function') {
-      return this.props.children(this.state.bootstrapped)
+    if (typeof children === 'function') {
+      return children(this.state.bootstrapped)
     }
 
-    return this.state.bootstrapped ? this.props.children : this.props.loading
+    if (this.state.bootstrapped) {
+      if (children) {
+        return children
+      }
+      return null
+    } else if (loading) {
+      return loading
+    } else {
+      return null
+    }
   }
 }


src/integration/react.js
```diff
@@ -51,16 +51,27 @@ export class PersistGate extends PureComponent<Props, State> {
...
+    if (typeof children === 'function') {
+      return children(this.state.bootstrapped)
     }
 
-    return this.state.bootstrapped ? this.props.children : this.props.loading
+    if (this.state.bootstrapped) {
+      if (children) {
+        return children
+      }
+      return null
```
erksch: This surely works! But in my opinion, this is what defaultProps should be used for.
You get around removing the `?` but it feels just like another workaround :D 
But the way flow handles defaultProps is to blame in the end.

erksch:
I like the destructuring though :D much cleaner to me.

src/integration/react.js
kareemsalah227: Yes I totally agree with you, I did mention that in the `important comment` section on the description, this should be Facebook's Flow work, not to be done here.

I'll try to take a look on Facebook's Flow repo and I might be able to make a good contribution.

kareemsalah227:
Good to hear, I'm happy to help :)

erksch:
> since this.props can't actually be set in the component code

Why not? This is what defaultProps are for. It overrides the optional props if they are not passed to the component.

kareemsalah227:
You're right, I take that back.

I just thought for a moment that this should be the parent component's responsibility to pass in default values for props, but that contradicts with the "optional props" concept.

You're right, ma bad, I owe you a :beer:

erksch:
No problem!
I did some further digging in https://github.com/facebook/flow/issues/1660 and found this really good comment https://github.com/facebook/flow/issues/1660#issuecomment-323898823.

The main argument is that looking at props from the components internal perspective, props that have a defaultProp are kind of not optional anymore (again from an internal perspective), because they always have a value and you don't have to recheck if they are set or not. Little bit weird but I get the point.

So removing optional flags seems to be the way to go.

kareemsalah227:
Agreed :)

aguynamedben:
I'm new to redux-persist and trying to help manage the project (it needs a lot more help if either of you is interested in helping). Thanks for both of your PRs.

So, if I'm following the discussion correctly does that mean https://github.com/rt2zz/redux-persist/pull/974 should be merged over https://github.com/rt2zz/redux-persist/pull/975?

1. I just merged this other PR to add a CHANGELOG to this project. Can one of you update the PR by merging in master (to get the CHANGELOG on your branch), then add a changelog entry describing this change? Does that make sense? Then I can merge it.

Thanks again for your contribution. If either of you wants to become a redux-persist expert we need more of those in the community! :)

aguynamedben:
Whoops, here's the changelog PR I mentioned: https://github.com/rt2zz/redux-persist/pull/954

erksch:
> So, if I'm following the discussion correctly does that mean #974 should be merged over #975?

Yep that's correct.
I'll follow your greatly numbered instructions!

erksch:
@aguynamedben My PR is ready now :)

erksch:
@kareemsalah227 If the change isn't too small for @aguynamedben I would encourage you to also contribute the object destructuring clean up.

kareemsalah227:
@erksch Sure thing, I'm currently out of home, will be back in a day or so, I hope this is not time critical.

aguynamedben:
Cool, I'll give @kareemsalah227 a few days to alter the PR with the object destructuring. No rush from my end. Thanks again, let me know when it's ready!

aguynamedben:
We merged https://github.com/rt2zz/redux-persist/pull/974 after discussion with @kareemsalah227 and @erksch, closing this PR.

PR #1002, opened 2019-03-11
Persist Store Callback

LorenGr:
When the store is persisted, a consumer would need to dispatch an action in his own app, he would therefore need a reference to the store he just persisted. This can be achieved simply passing it through the persist store callback.

Implementation example :
persistStore(store,[null, store => store.dispatch({type:"PERSONAL_ACTION"})]);


reinierv4:
Anyway to pass just the action as an additional argument to the persistStore instead?

rt2zz:
you already have store in scope, why not just use that?
```
persistStore(store,null, () => store.dispatch({type:"PERSONAL_ACTION"}));
```

rt2zz:
closing, please let me know if I am missing something with `persistStore(store,null, () => store.dispatch({type:"PERSONAL_ACTION"}));`
`

PR #1006, opened 2019-03-16
update asyncstorage access , 

karkipy:
retrive from @react-native-community/async-storage
@pascience @af @ssorallen @joeybaker


iagorm:
Just waiting this merge to upgrade the dependency, good job!

Ilphrin:
Thanks for this PR, I hope it will be merged and released soon =D

bartekus:
Why isnt this merged yet?
```
Warning: Async Storage has been extracted from react-native core and will be removed in a future release. It can now be installed and imported from '@react-native-community/async-storage' instead of 'react-native'. See https://github.com/react-native-community/react-native-async-storage
```
Given the above, as well as this PR's size and specificity at addressing this issue, this PR should be merged asap.

mazenchami:
@aguynamedben can you help with this PR?

package.json
```diff
@@ -85,7 +85,8 @@
     "storage-memory": "0.0.2"
   },
   "peerDependencies": {
-    "redux": ">4.0.0"
+    "redux": ">4.0.0",
+    "@react-native-community/async-storage": "^1.2.1"
```
chris-feist: This should be an [`optionalDependency`](https://yarnpkg.com/lang/en/docs/dependency-types/#toc-optionaldependencies) for those that aren't using this with RN

bartekus: But the only thing that optionalDependencies is good for is enabling an install to not fail if that dependency does not successfully install! One important thing to note here is that it will still attempt to install these optionalDependencies.

On the other hand, regarding peerDependencies, with npm 2, if your package.json listed a peer dependency, when someone installed your package, that dependency was also automatically installed but with npm 3+/yarn installation no peerDependency will actually be installed and instead a warning message about missing peerDependencies will be show.

As such, I believe that moving `react-native-community/async-storage` from peerDependencies to optionalDependencies will actually cause the opposite effect to what your proposition intends.
At worst with `peerDependencies` the non-RN users will see a warning about `redux-persist` missing `@react-native-community/async-storage` peerDependencies.
At worst with `optionalDependency` the non-RN users will end up with `node_modules` that actually includes `@react-native-community/async-storage` package.

Yarn perspective:
[peerDependencies](https://yarnpkg.com/lang/en/docs/dependency-types/#toc-peerdependencies)
[optionalDependencies](https://yarnpkg.com/lang/en/docs/dependency-types/#toc-optionaldependencies)

npm perspective:
[peerDependencies](https://docs.npmjs.com/files/package.json#peerdependencies)
[optionalDependencies](https://docs.npmjs.com/files/package.json#optionaldependencies)

related articles:
[common-npm-mistakes](https://medium.com/@jacob.h.page/common-npm-mistakes-51bf8989079f)
[understanding-the-npm-dependency-model](https://lexi-lambda.github.io/blog/2016/08/24/understanding-the-npm-dependency-model/)

rlemasquerier:
Hi! Any idea on when this PR will be merged?

package.json
AleksandrChernyavenko: Maybe it should not be in package dependencies at all because to make it works you still need to link package manually `react-native link @react-native-community/async-storage`.

Instead we can add  info about usage of `@react-native-community/async-storage` to README.md (seems we already have it)

rt2zz: I agree this should not be in deps at all. Instead we should remove `index.native.js` and update the readme accordingly.

rt2zz:
I dont want to add this to deps, so I propose we simply remove `index.native.js` update the docs accordingly (users should import and use AsyncStorage directly) and publish it as a v6.

rt2zz:
https://github.com/rt2zz/redux-persist/releases/tag/v6.0.0-pre1

Now removed built in for AsyncStorage, users will need to do `import AsyncStorage from '@react-native-community/async-storage';` (after linking). This will solve dep / linking issues as well.

Meph1k:
Thanks!

rt2zz:
I should mention its behind a next tag, easiest way to get the latest is `npm i redux-persist@next`

PR #1035, opened 2019-05-15
Remove useBuiltIns from babelrc

codebutler:
Fixes issue with `6.0.0-pre1` where every app using this library must manually add a dependency on `core-js@2`.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,6 @@
   "presets": [
     "@babel/preset-flow",
     ["@babel/preset-env", {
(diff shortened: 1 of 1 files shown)


chrisabrams:
Out of curiosity why is `useBuiltIns` being used?

rt2zz:
im open to removing this but am not up to speed on the state of babel. If we remove this will there be more polyfills or less? if there are less there might be an issue with our usage of async await and object spread.

PR #1047, opened 2019-06-04
Something about code splitting

mrtrebor:
It would be great to add something about code splitting if redux-persist has to be used. **Don't use this code**.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ Usage Examples:
 1. [Basic Usage](#basic-usage)
 2. [Nested Persists](#nested-persists)
 3. [Hot Module Replacement](./docs/hot-module-replacement.md)
-4. Code Splitting [coming soon]
+4. [Code Splitting](./docs/code-splitting.md)
 
 #### Basic Usage
 Basic usage involves adding `persistReducer` and `persistStore` to your setup. **IMPORTANT** Every app needs to decide how many levels of state they want to "merge". The default is 1 level. Please read through the [state reconciler docs](#state-reconciler) for more information.
--- a/docs/code-splitting.md
+++ b/docs/code-splitting.md
@@ -1,0 +1,51 @@
+## Code splitting
+
+With code splitting, we can load only needed in given moment chunks of code used by an application.
+
+We already used `replaceReducer` when we talked about HMR.
+
+This function is required to replace old reducer with newly created reducer which includes dynamically injected ones.
+
+Code splitting is described well at https://redux.js.org/recipes/code-splitting.
+
+**configureStore.js**
+```js
+import { persistReducer } from 'redux-persist'
+import rootReducer from './path/to/reducer'
+
+// Static reducers...
+
+export default () => {
+  // ...
+
+  const store = createStore(
+    // createReducer returns static reducers combined with injected ones
+    createReducer({ /* here without async (dynamic) reducers */ }),
+    initialState,
+    composeEnhancers(...enhancers),
+  );
+
+  // Async reducers registry (in react-boilerplate they are called injected reducers)
+  // Here we track dynamically injected reducers
+  store.asyncReducers = {};
+
+  const persistor = persistStore(store, null, () => {
+    // Get persisted state
+    Object.keys(store.getState())
+      // We always have static reducers loaded
+      .filter(reducerKey => !Object.keys(staticReducers).includes(reducerKey))
+      .forEach(reducerKey => {
+        // Create empty reducers for keys that don't have loaded dynamic reducer yet
+        // They will be replaced be a real ones.
+        store.asyncReducers[reducerKey] = (state = null) => state;
+      });
+  });
+  store.persistor = persistor;
+
+  // Some method like injectReducer which adds async reducer to registry...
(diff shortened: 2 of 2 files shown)


abdullah-almesbahi:
dear @mrtrebor , when I use your code ,connected-react-router stop changing URL when navigate to any page.

Do you use connected-react-router in your project?

abdullah-almesbahi:
I fixed like this
```
  const persistor = persistStore(store, null, () => {
    // Get persisted state
    Object.keys(store.getState())
      // We always have static reducers loaded
      .filter(reducerKey => !Object.keys(createReducer()).includes(reducerKey))
      .forEach(reducerKey => {
        // ignore router
        if (reducerKey !== "router") {
          // Create empty reducers for keys that don't have loaded dynamic reducer yet
          // They will be replaced by a real ones.
          store.injectedReducers[reducerKey] = (state = null) => state;
        }
      });
  });
```

mrtrebor:
I prefer instead to put a router key into the blacklist of persistReducer's config.
```
const rootPersistConfig = {
  key: 'root',
  storage,
  blacklist: ['router'],
};
// ...
const reducers = persistCombineReducers(rootPersistConfig, reducers);
```

abdullah-almesbahi:
@mrtrebor btw, this make the reducer get injected
```
 useInjectReducer({ key: "profile", reducer });
```

but dose not persist data from store. always return initial data

is it working for you?

abdullah-almesbahi:
@mrtrebor another error found. when run logout action , used this action in many reducer to reset initial state on logout action . only works with current injected reducer , but other static reducer doesn't do any thing

mrtrebor:
@abdullah-almesbahi I don't use `useInjectReducer`. I can't help you without a code in some online sandbox as an example of your problem. That errors can be related to your implementation.

mrtrebor:
I had some struggles with an app using the previous solution, so I ended with a better approach. The code in the latest commit.

joshmeads:
@mrtrebor nice work, thanks for this!

One small question though, do you get a large amount (3-6 for me) of "persist/REHYDRATE" actions firing with "Error: redux-persist: persist timed out for persist key <key>" after the initial successful one?

I've tried adding debounce and timeout into the config as other issues suggest with no success.

```js
// Persisted redux config
export const persistedConfig = {
  key: 'persistedRedux',
  storage: localForage,
  blacklist: ['router'],
  keyPrefix: '',
};
```

```js
// reducers.js
export default function createReducer(injectedReducers = {}) {
  const reducers = {
    router: connectRouter(history),
    ...globalReducers,
    ...injectedReducers,
  };

  return persistCombineReducers(persistedConfig, reducers);
}
```

```js
// configureStore.js
  const staticReducerKeys = ['router', '_persist', ...Object.keys(globalReducers)];
  const dumbReducers = Object.keys(initialState)
    .filter((reducerKey) => !staticReducerKeys.includes(reducerKey)) // We always have static reducers loaded
    .reduce((result, reducerKey) => {
      // Create empty reducers for keys that don't have loaded dynamic reducer yet
      // They will be replaced by the real one
      result[reducerKey] = (state = {}) => state;
      return result;
    }, {});

// ... 

 const store = createStore(
    createReducer(dumbReducers),
    {},
    composeEnhancers(...enhancers)
  );

  store.injectedReducers = { ...dumbReducers }; // Reducer registry

  store.persistor = persistStore(store, null);
  store.persistor.persist();

  if (module.hot) {
    module.hot.accept('./reducers', () => {
      store.replaceReducer(createReducer(store.injectedReducers));
      store.persistor.persist();
    });
  }

  return store;
}
```

mrtrebor:
@joshmeads Hi, I'm receiving multiple REHYDRATE calls and I think it's related to injecting reducers (which are loaded async and state need to be rehydrated). but it works for me without error. Did you try to use another storage engine?

joshmeads:
> Did you try to use another storage engine?

Yes, I tried the default localstorage too. I'll play around with it further and post any findings if I manage to solve it. Cheers

ygr1k:
Pls merge this into the doc, it's really helpful.
@mrtrebor thanks for your work!

pedrosimao:
I have used this code example on my app, but whenever Reducer Injections happens my Redux Store gets a new `@@INIT` action and I can no longer see the history of all actions on redux-debugger. I also have multiple `REHYDRATE` actions. Is that normal or is there a way to inject reducers without loosing the redux actions history?

mrtrebor:
@pedrosimao I think you could try to change redux-devtools-extension's parameter [shouldHotReload](https://github.com/zalmoxisus/redux-devtools-extension/blob/master/docs/API/Arguments.md#shouldhotreload) to **true**.

pedrosimao:
@mrtrebor Thanks a lot for your support. I have tried changing `shouldHotReload` to `true` as suggested, but I still get `@@INIT` actions and the history completely erased. I also feel there are more `persist/REHYDRATE` actions than before, but not sure about that. The only way I managed to correct the problem is to completely abolish reducer injections. But this is hard, since the project has a lot of sub-components using the function...

PR #1073, opened 2019-08-19
Move @types/react to dependencies.

tai2:
Because @types/react is placed in devDependencies, I cannot have @types/redux-persist-expire test passed.
Why it can't? My package depends on a type which redux-persist exposes, so it declares [redux-persist as a dependency](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7f128858b9f1556567877d678945c8e3d757a807/types/redux-persist-expire/package.json#L5). 
When DefinitelyTyped's CI runs, it installs dependent packages then [run tests and fails](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/573622586?utm_source=github_status&utm_medium=notification). 
@types/react won't be installed because it is declared in devDependencies.

As well as source code itself, @types dependencies should also be installed when install the library.


wouterds:
Just install it as a dependency in your project, it's not a dependency of this project.

tai2:
I can't install it in my project which is a DefinitelyTyped's subproject.
Because DT doesn't permit install it in devDependencies.

tai2:
Further discuttions: 

* [Saving npm @types typings with --save or --save-dev](https://stackoverflow.com/questions/40097488/saving-npm-types-typings-with-save-or-save-dev)
* [Why --save instead of --save-dev? #81](https://github.com/microsoft/types-publisher/issues/81)

devDependencies is preferred way in our case as long as I understand it.

tai2:
In addition to what I mentioned above, DT doesn't permit me to import what I use directly.
@types/react is an indirect dependency to my package.

Issue #1092, opened 2019-09-20
PersistConfig props missing from Typescript definitions


aschultz:
Looks like a few of the newer options on PersistConfig are missing from the TS definitions (serialize, deserialize, and writeFailHandler).

PR #1093, opened 2019-09-23
chore: add missing type for persist config

Naturalclar:
This PR addresses #1092 
Added deserialize type definition and added function for serialize type

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -42,7 +42,8 @@ declare module "redux-persist/es/types" {
      */
     getStoredState?: (config: PersistConfig<S, RS, HSS, ESS>) => Promise<PersistedState>;
     debug?: boolean;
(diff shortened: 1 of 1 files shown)


types/types.d.ts
```diff
@@ -42,7 +42,8 @@ declare module "redux-persist/es/types" {
      */
     getStoredState?: (config: PersistConfig<S, RS, HSS, ESS>) => Promise<PersistedState>;
     debug?: boolean;
-    serialize?: boolean;
+    serialize?: boolean | (data: any) => any;
+    deseriazlize?: boolean | (serialized: any) => any;
```
romainbraun: typo

Naturalclar: oops, thanks for the review!
I had it fixed but I forgot to push the commit 😅

PR #1153, opened 2020-01-30
fix(types): add deserialize option to config

jaulz:


toytonic:
anything missing here? was about to open the same PR ;)

PR #1185, opened 2020-04-07
TypeScript type definition update

zoltan-boros:
Field manualPersist added to PersistorOptions in TypeScript type definitions.


nickradford:
@rt2zz can this get reviewed / merged / released?

Issue #1210, opened 2020-06-09
Using non oppressive language


martymclaugh:
I'd like to request that this project start using non oppressive language and switching variables like `whitelist` and `blacklist` to `allowList` and `blockList`

lumberman:
1936

leandrojo:
I also came here thinking about the same thing. I think that **include** and **exclude** are good terms.

pro-nasa:
It's not oppressive. It never applied to race, like bright stars in the black sky or white-black chess. The blacklist is colloquial speech. For example "you are on my blacklist". This is how freedom of speech in Western civilization dies

pro-nasa:
the worst part is that instead of focusing on the real problem of racism, people only make it deeper in this way

pro-nasa:
The view that terms such as “blacklist” and “whitelist” reinforce the notion that black==bad and white==good it's highly illogical because is out of context. The world has gone crazy. And I know that nothing can stop it. The rule of idiocracy begins

pro-nasa:
BTW. I don't mind replacing to allowlist and blocklist, only that it is done for ideological reasons. This is pure communism

tcoughlin3:
If there's even a small chance that someone would think a naming convention is rooted in racism, then just change it regardless of whether you feel it's logical. It's just a search and replace.

pro-nasa:
https://github.com/rt2zz/redux-persist/pull/1211
https://github.com/rt2zz/redux-persist/pull/1213
https://github.com/rt2zz/redux-persist/pull/1242
https://github.com/rt2zz/redux-persist/pull/1289

PR #1211, opened 2020-06-10
change variables to use non oppressive language

martymclaugh:
change whitelist to allowList and blacklist to blockList


lumberman:
Do you realize that by simply renaming config keys you will break all the apps using this library?

zilahir:
oh god :(

vincenzovitale:
@lumberman and who cares? Let's change those libraries as well with PRs, forks or whatever.

Language and words are the first things requiring evolving for other changes to happen!
Blacklist, master, slave and so on represent a bad past and good to start flagging those terms.

JacobJaffe:
> Do you realize that by simply renaming config keys you will break all the apps using this library? 
Let's figure out a constructive path forward then

Can these keys retained, but deprecated? 
Can we introduce the new keys alongside the old for now, and on a next breaking version update these would be the default?

leandrojo:
@lumberman

MAJOR version when you make incompatible API changes;
https://semver.org/

nromptea42:
You should not push the `yarn.lock` file if there is already a `package-lock.json`

lumberman:
> @lumberman and who cares? Let's change those libraries as well with PRs, forks or whatever.

I care. This change will break the final applications, not other libraries that you can patch.

lumberman:
> Can these keys retained, but deprecated?

How that "eliminates oppressive language" in the code? I will still see that horrible blacklist in the source code! 🙀

tincho:
what about stop calling "black" people when their skins are actually more in the shade of brown
and well, same for "white" people, who have kind of light-pink skin

There, and not here (in variable names like blacklist) is the bug

rossPatton:
`allowlist` and `blocklist` is clearer language anyway. just deprecate white/black for a few releases, with a warning, and then transfer over to new keys? maybe i'm missing something but it doesn't seem like a huge change to me?

martymclaugh:
@lumberman you will also see how we made an effort to improve things instead of doing nothing.

@tincho it's too bad we can't revert that commit on humanity's repo: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/

How about let's figure out a way to make this work instead of arguing against it? 

@leandrojo would updating the version to 7.0.0 make it a nonbreaking change? https://semver.org/

tincho:
let's put it this way:
some people think blacklist and whitelist is "oppresive" and racist 
other people don't think so (myself included) 

at least we should run a poll to see how many are in each group, worldwide, and by country

proposing changes is OK to me (if you think there will be less opression in the world after replacing words in thousands of files you are free to believe it)
questioning what's established, and trascend the "black-white" metaphor is OK too

but say it's oppresive and racist is fallacy 
the link you cite does not explain a clear link between the dualism and the races...only throwing labels left and right and stating "its racist and must stop". no clear argument

rossPatton:
hmm, how is it not clear? 

black in this context === bad
white in this context === good

master === owner
slave ==== you get the picture

It's not a complicated thing that's being proposed here. No one claims that adjusting language to match the changing norms of the times is going to magically wipe away oppression. Seems like you are arguing against a strawman of your own creation.

It's a change that makes the language clearer. `allow` and `deny` are just more clear terms to use, regardless of how you feel about the politics of the issue.

It costs next to nothing to change, might make this project feel more welcoming to people, and implements terms that make the codebase more clear.

Should be cut and dry, imo

tincho:
Agree with most of what you say. Would agree to merge this PR anyway, were I the person to decide it. For sure it would have more positive reactions than "doing nothing" and I would celebrate that.

I just dont believe that black-white dualism is related nor based on human races. This interpretation seems to be more based in the socio-emotional context of this time that in any other reason. I judge it as a legitimate reason, I just don't agree with it. 

[The metaphor is wider in meanings than that...  been around for thousands of years and they say it comes from Middle East, where I don't think they had too much white privilege by that time](https://en.wikipedia.org/wiki/Black-and-white_dualism) )

Wanted to make my point because respectfully interchanging different views is good habit to me. One can always be "wrong".

And because IMHO, accepting to call oppresive or racist to things that are not (still havent got any logical argument for it) has the backfiring effect that got [the boy who cried wolf](https://fablesofaesop.com/the-boy-who-cried-wolf.html) eaten. Normalizing to accept subjetive interpretations as "the obvious truth" is risky. Everything must be healthily questioned. Maybe one day some will say "heck, everything is racism nowdays, so who cares"
This is my - I hope - constructive criticism
Best regards to all

bfunk88:
Breaking changes are to be expected from time to time. As long as the version is bumped appropriately, I don't see the problem at all. It's ultimately a developer's responsibility to scan change logs for breaking changes when installing a major(ish) package upgrade.

schumannd:
@bfunk88 there should be a period of deprecation, however. Introducing a breaking change without a transition period is problematic.

pro-nasa:
@rossPatton https://github.com/rt2zz/redux-persist/issues/1210#issuecomment-664837484

pro-nasa:
@lumberman It's something like backward compatibility. I don't mind this change but with keeping old keys for a while

alexdroste:
I don't think that this is the right place for politics.
Whether you like the naming or not from a technical point of view it is absolute madness to break hundreds and thousands of apps because someone wanted to rename a public interface variable name.

This pull request is hypocrisy at its finest.

pro-nasa:
@alexdroste Such changes are lobbied at the very levels of governments/companies (this is followed by regulations) so it will have to be changed for the sake of peace of mind

256hz:
If this pull request is only lip service, why not pay it?  It's a tiny price for improving the language as a whole.

Breaking apps should not be a concern.  This is what semantic versioning was invented for.  Increment the minor or even major version, if that is at the root of your fear.

Of course this is the place for politics.  Everything is political, whether you choose to acknowledge it or not.  No single action will fix inequality - it will be the sum of millions of tiny actions.  Why not this one?

pro-nasa:
@256hz Agree, it's good we're talking about it. That is the purpose of the discussion. It would be good if @rt2zz referred to it.

@rossPatton Personally, when I try to rationalize it, I don't agree that blacklist/whitelist being offensive as long as not applicable to humans/race. I also disagree that black=bad and white=good in this context. Both are neutral in my opinion. There may be very important and sensitive things in a blacklist that we do not want to store for security reasons, or there may be things considered irrelevant. It's just a list of keys that will be kept / not kept. And nothing more.

Perhaps in this case it's not worth rationalizing the change itself, but rather reflecting on its main goal. Like showing empathy to people who are sensitive / have experienced discrimination. After all, does language change as culture changes? I didn't understand it at the beginning, because I'm an insensitive man who hasn't experienced discrimination and lives in a fairly homogeneous group of people.
I think @papb put it nicely here: https://github.com/rt2zz/redux-persist/pull/1242#issuecomment-698041922

pro-nasa:
@256hz wrote:
> Of course this is the place for politics. Everything is political, whether you choose to acknowledge it or not.

I think the cause of all the action in public today is: Herbert Marcuse "A Critique of Pure Tolerance"

256hz:
Thanks for weighing in @pro-nasa, I'm happy to hear you are still considering this change.  I agree strongly with the comment you linked.  I'm always a surprised at the strong reaction some have to this group of code language changes. Maybe they are performative, or useless, or silly; then why the violent opposition?  These changes are small and we can do them and it's possible that we just might help one person one day.  I'm good with that.

pro-nasa:

tincho:
> Thanks for weighing in @pro-nasa, I'm happy to hear you are not so strongly opposed. I agree very much with the comment you linked. I'm always surprised at the reaction to this group of code language changes. Maybe they are performative, or useless, or silly; then why the violent opposition? They are small and we can do them and it's possible that we just might help one person one day. I'm good with that.

from my part there is no opposition, much less violent. 

I just need to point out that in no way these terms in these places were racist at all. In no way they are oppresive or offensive or violent. 

There is no "good" or "bad", this is objective. And it's **not** about people, or people's skin color or whatever else personal trait. 

A "black list" is a list that happens to use the color "black" meaning "negation", because, darkness is in essence absence of light. 

Which, if we speak of skin colours, is the exact opposite, "white" people lack some pigmentation that "black" people do have. BTW there's not a single peson in the world that has either `#000` black nor `#FFF` white skin....at best we are pink-ish or brown-ish... 

If we start to accept there's racism (or whetever else hateful or violent attitudes) where __there isn't__, then we'll fail to discern where it _really_ is. That's why this matters to me, I don't care about "breaking changes in code". 

I understand the intentions but I see the measure as counter-productive. You know the saying, "the road to hell is paved on good intentions"...

256hz:
I'm not sure I understand.  You open with

> from my part there is no opposition...

but your whole response is about why it shouldn't be done...

You seem to think your view is 'objective', or inarguable, and that you can speak for everyone.  I don't agree.

Any word can become loaded.  "Blacklist" is 400 years old, and what it means to you or me, now, is going to be from the viewpoint of a different culture.  Neither one of us can say what it means to everyone.  There is no 'objective' truth in cultural connotations of language.

So, I can't say definitively that these terms **are** racist.  But in an industry that's majority white, and where black people are under-represented by population, it's not a stretch to say we could be more inclusive.  Is changing "blacklist" and "whitelist" going to revolutionize the tech industry?  Of course not.  But there is the chance that it helps one person feel more included when using this library.  And does it take any effort on our part?  No, besides convincing the maintainers to make the change.

I'll ask you this: what do you stand to gain by not changing the language?  Is it worth possibly alienating that future developer?  It doesn't seem possible to me that it is.

tincho:
I have no opposition to the change. I have a strong argument behind it's alleged reasons. 
Change the language, change the words, go ahead. I'll be chill as a hill.
 
I'm saying: don't fool yourselves believing this makes anything any more inclusive. Don't try to force the idea that this wording may alienate somebody. It doesn't. People will get offended by whatever they want. That's a problem of their own.

I'll give you an example. I'm left-handed, so that makes me part of a minority ( :partying_face: ¿?  ) across all races, genders, cultures, whatever. We are "under represented" almost everywhere. The words used to call us in some language are even insulting (see: gauche in french, sinister... etc), while the words for "right-handed" are quite virtuous (see: dextrous). Even "right" means something "good" (see: human rights, righteous, etc) and "left" is... well, what's left, leftovers, etc.

Do I feel oppresed? Of course not! I don't give two chodes about it. 
Could I say: "I feel oppresed"? Of course. ~It's the fashion nowadays~ 
I could even state that many left handed were oppresed and killed in the past. 
Was I? Nope I wasn't!!
Will it do me any service, to play victim? None at all. That's why I choose not to feel oppresed, and there: problem gone!
Will my life be any better if those words change? In no practical way. 
Is this alienating me? No, it isn't!
But if I were to convince enough people that it is, and start a movement to cancel such words, there would actually be 2 real outcomes:

1. No perceivable changes in real life
2. LESS FREEDOM

Don't you see it? Banning "bad" words is what [newspeak](https://en.wikipedia.org/wiki/Newspeak) was all about: to prevent people from thinking in certain ways, thus forcing them into others.

My point is: everyone can choose how to feel about something. We can't prevent it in no way. And me, you, or anybody,  don't need to advertise how non-racist we are by removing words "just in case anyone may find it offensive". That's a decision of whoever interprets it. 
What's happening here is that some are saying: "interpret these words this way, they are evil, they are racist, etc". That's underestimating the audience's intelligence.

So yes, there are fewer black programmers than white ones. Could anyone be naive enought to think it has something to do with the usage of words like `blacklist` or `master/slave`? There are deeper reasons than that!!

We're missing the point here. We start with little changes like these ones, advertised as "being more inclusive", and next thing we know we get more and more words banned, replaced with euphemisms, and we'll get used to it. 
Racism, hate, and any societal issues **will still be there**, but we'll have... **less freedom**! All of us, and first of all, those who are oppressed.

So, I insist: go change any word you want. If you think about it for a moment you'll see we've been actually doing it for years. And problems do not decrease, they still increase. It's madness to expect different outcomes doing the same thing over and over.

256hz:
So why not use racial slurs that were common in past times then?  Your "freedom" is worth more than anyone else's comfort, right?  It's not 1984 to stop using words that offend people.  If you really think "stop feeling oppressed" is the solution to oppression, you have no idea what oppression means.

Problems have increased since it fell out of fashion to call people the n-word?  I can't understand any, literally any, of the conclusions you've drawn here.  I have no doubt you think your conclusions are the "logical, objective" way of thinking, but I have to tell you, nothing you've said makes any sense to me, logically or otherwise.

tincho:
These are my personal views, backed by facts. I'd rather let the reader tell which is which.

It's **our** freedom. All of us, no matter if offended or unoffended.

Words don't offend people, people offend people (and often we find ourselves offended by pure choice)

You're confusing "oppress" with "offend" which is far very different. I'd say it's kind of disrespectful for people who's suffered real oppresion, and I mean, not by some words in source code.

The  n-word has been avoided for years, still today in 2021 we see racism and oppression. That kind of proves my point even more. Many people seem to have not realised it happened until 2020, when BLM got great visibility, but guess what? it happened before, and still happens.

Anyway, thank you for your views, as I said, I share mines not from a "conservative" stance, more of a "DUDE, SRSLY?" point of view.

PR #1213, opened 2020-06-17
(persistConfig) change whitelist to allowlist and blacklist to blocklist

leandrojo:


bitflower:
Exactly my thoughts right now!

tindn:
thank you for submitting this PR

dbarner1:
👏

martymclaugh:
Had I missed something in #1211 ?

DaC00P:
Can we get this merged? The org I work for doesn't want to keep using the packager otherwise. Updated would be appreciated @rt2zz

lisamancke:
Thank you for this PR.

pro-nasa:
political correctness will kill us all (only because someone has associated the whitelist and blacklist with racism)

NearHuscarl:
This is one of the most stupid PR I've ever seen alongside with the santa denier PR from vscode. If it works, don't change it just because you think it's cool to be trendy and make life harder for anyone else. The concept of whitelist and blacklist has been there forever and the overwhelmingly majority of the people do not connect it with racism at all. You are trying to solve the problem that simply doesn't exist.

gabmmontenegro:
Simple but important PR to avoid offensive terms and promote inclusive language. It doesn't make sense just for those who don't understand structural oppression.

**Recommendation:** denylist/allowlist

Motivations for change:
[Linux kernel will no longer use terms ‘blacklist’ and ‘slave’](https://thenextweb.com/dd/2020/07/13/linux-kernel-will-no-longer-use-terms-blacklist-and-slave/
)
[Tech companies update language to avoid offensive terms](https://www.marketplace.org/2020/06/17/tech-companies-update-language-to-avoid-offensive-terms/)

pro-nasa:
`denylist` sounds better than `blocklist`

pro-nasa:
@gabmmontenegro So the devil is white and the angel is black? The bride in black for the wedding, in white for the funeral?

gabmmontenegro:
> @gabmmontenegro So the devil is white and the angel is black? The bride in black for the wedding, in white for the funeral?

This PR is making us reflect on that and that's a good. The question we should always ask is "why?". Go for it. Just consider that words and symbols shape the dominant culture making oppression almost invisible even to the oppressed ones.

pro-nasa:
The point is, it has never been oppression until now. It has never been related to race. Human - as a daytime animal - has always been afraid of the dark. In all cultures light (white) and dark (black) have always been understood in the same way. Even for cultures whos never seen a black/white man. Some people are trying to give new meanings to words that were created on this basis. And why? Because some countries have shame history and the problem of racism elevated to the level of absurdity? Why is it so hard for BLACK and WHITE people to understand? It makes that a lot of people like me feel oppressed because must participate in a comedy about creating problems that were unknown before.

But as I said in another post I'm not going to block the renaming if others want it (I have no such power to do this). I just disagree with the arguments for which it is being done.

pro-nasa:
And no, just because big tech companies do it doesn't mean they are right. They always want to be warriors to change the world

pro-nasa:

Issue #866, opened 2018-07-12
Enable `createMigrate` to handle async migrations


ssorallen:
The function passed to `migrate` can already handle a `Promise`, but `createMigrate` expects each of the integer migrations to be synchronous, simple synchronous transforms of PersistedState => PersistedState. It would be great to enable serial, asynchronous migrations in `createMigrate`. The change could be localized to `createMigrate` since the rest of of redux-persist already expects `migrate` to be asynchronous.

aguynamedben:
To deal with this, I'm copying this project's createMigrate.js and modifying it in my project to this: https://gist.github.com/aguynamedben/ee5ef358856f1543129265fd3cf8b732

See [the tests](https://gist.github.com/aguynamedben/ee5ef358856f1543129265fd3cf8b732#file-customcreatemigrate-spec-js-L4) for how to use it...

```js
      const asyncMigrations = {
        1: async (state) => {
          return {
            ...state,
            foo: 'different',
          };
        },
        2: async (state) => {
          return {
            ...state,
            baz: 'bang',
          };
        },
        3: async (state) => {
          return {
            ...state,
            cool: 'beans',
          };
        },
      };
```

configureStore.dev.js
```js
  const persistConfig = {
    key: 'root',
    storage: storage,
    debug: true,

    migrate: customCreateMigrate(asyncMigrations, {
      debug: true,
      asyncMigrations: true,
    }),
    version: 8,
  };
```

We use the same logic in customCreate.js to figure out which migrations to run. For async migrations, we use [this technique](https://decembersoft.com/posts/promises-in-serial-with-array-reduce/) to dynamically build a promise chain using Array.reduce(). This means the async migrations are run in sequence, one after another.

With the current code, you must provide the option `{ asyncMigrations: true }` or else the old logic is used, which will screw up if you have async migrations.

If #929 gets resolved I'd happily submit a PR to add this. I think we could even detect if the migration function is async and resolve automatically so that specifying the `asyncMigrations` option isn't necessary.

aguynamedben:
I need to submit my gist as a PR.

aguynamedben:
I'm not sure if this is also related to feature requests for async transforms. https://github.com/rt2zz/redux-persist/issues/303 I need to figure that out. It seems like they might be separate per this comment: https://github.com/rt2zz/redux-persist/issues/426#issuecomment-325262469

Also maybe related: https://github.com/rt2zz/redux-persist/issues/426

[...]

Issue #806, opened 2018-04-23
Redux-Persist v5 migrate from one storage system to another


alanlanglois:
I'm working on a project where I need to migrate my data initially stored using AsyncStorage to another one: redux-persist-fs-storage

Based on this issue: https://github.com/rt2zz/redux-persist/issues/679 I've manage to get the data stored in AsyncStorage this way:

```
let persistor = persistStore(store, persistConfig, async () => {
  try {
    console.log("GET ASYNC STATE")
    const asyncState = await getStoredState({
      key: 'root',
      storage: AsyncStorage,
     })
     console.log("ASYNC >>>> " + asyncState )
       if (asyncState) {
     }
   } catch (error) {
    console.warn('getStoredState ASYNC error', error)
  }
});
```
I'm looking a way to inject these data (asyncState) into the new fileSystem store.

In v4 we could achieve that using a rehydrate method on the persistor.

`fsPersistor.rehydrate(asyncState, { serial: false })`
It's not a migration from v4 to v5, just from a store to another using v5.

Can't find a way to do it in v5. An input on how to do it would be great :)

Cheers

hutchy2570:
Hi,

I've managed to get this to work using v5's migrate feature.

`isStateEmpty` is a function which you craft to determine whether the state is empty for your use case.

```
const migrate = async state => {
  // Migrate from async storage to fs https://github.com/robwalkerco/redux-persist-filesystem-storage#migration-from-previous-storage
  __DEV__ && console.log('Attempting migration');
  if (isStateEmpty(state)) {
    // if state from fs storage is empty try to read state from previous storage
    __DEV__ && console.log('FS state empty');
    try {
      const asyncState = await getStoredState({
        key: 'root',
        storage: AsyncStorage,
      });
      if (!isStateEmpty(asyncState)) {
        __DEV__ && console.log('Async state not empty. Attempting migration.');
        // if data exists in `AsyncStorage` - rehydrate fs persistor with it
        return asyncState;
      }
    } catch (getStateError) {
      __DEV__ && console.warn('getStoredState error', getStateError);
    }
  }
  __DEV__ && console.log('FS state not empty');
  return state;
};
```

You then just include this function as the `migrate` property of the PersistConfig passed to `persistReducer`

forster-thomas:
@hutchy2570 where I put this code?

piotr-cz:
This is my take on migrating from custom (legacy) storage

```js
// configureStore.js
[...]

PR #1219, opened 2020-07-01
Enable `createMigrate` to handle async migrations #866

retyui:
via: https://github.com/rt2zz/redux-persist/issues/806

```js
const migrations = {
  0: (state) => {
    // migration clear out device state
    return {
      ...state,
      device: undefined   
    }
  },
  // You can use also `async () => {}` function
  // For example to move users to other storage
  1: async ({ users, ...state }) => {
    try {
      await insertToOtherStorage(users)

      return state
    } catch (e) {
      // Don't forget about exceptions
      return state
    }
  },
  2: (state) => {
    // migration to keep only device state
    return {
      device: state.device
    }
  }
}
```

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -13,7 +13,19 @@ const migrations = {
       device: undefined   
     }
   },
-  1: (state) => {
+  // You can use also `async () => {}` function
+  // For example to move users to other storage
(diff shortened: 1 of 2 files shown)


AlimovSV:
Any progress?

retyui:
cc @ckalika

PR #1230, opened 2020-07-22
fix(types): Add manualPersist to PersistorOptions

nickradford:
This is a missing type in the `types.d.ts` file


nickradford:
Closing in favor of #1185

PR #1240, opened 2020-08-14
Move sealed variable outside middleware

magrinj:
This fix an issue when you use `replaceReducer` and `persist` right after in debug mode.
When you're in debug doing the above actions will call the rehydrate function in the timeout due to the call to `replaceReducer` and emit a timeout error.
Keeping variable outside sub middleware and initialising it at each `PERSIST` action fix the issue.


Energizz:
+1

PR #1242, opened 2020-08-28
Terminology, Power and Oppressive Language

sanntanna:
The central objective of this is to no longer use terms that no longer fit to be used within an inclusive culture and seeking diversity. We could help the community as a whole to find less aggressive ways to convey some concepts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ const App = () => {
   - arguments
     - [**config**](https://github.com/rt2zz/redux-persist/blob/master/src/types.js#L13-L27) *object*
       - required config: `key, storage`
-      - notable other config: `whitelist, blacklist, version, stateReconciler, debug`
+      - notable other config: `allowedlist, avoidlist, version, stateReconciler, debug`
     - **reducer** *function*
       - any reducer will work, typically this would be the top level reducer returned by `combineReducers`
   - returns an enhanced reducer
@@ -136,26 +136,26 @@ Redux persist ships with react integration as a convenience. The `PersistGate` c
 1. `loading` prop: The provided loading value will be rendered until persistence is complete at which point children will be rendered.
 2. function children: The function will be invoked with a single `bootstrapped` argument. When bootstrapped is true, persistence is complete and it is safe to render the full app. This can be useful for adding transition animations.
 
-## Blacklist & Whitelist
+## AvoidList & Allowedlist
 By Example:
 ```js
-// BLACKLIST
+// AVOIDLIST
 const persistConfig = {
   key: 'root',
   storage: storage,
-  blacklist: ['navigation'] // navigation will not be persisted
+  avoidlist: ['navigation'] // navigation will not be persisted
 };
 
-// WHITELIST
+// ALLOWEDLIST
 const persistConfig = {
   key: 'root',
   storage: storage,
-  whitelist: ['navigation'] // only navigation will be persisted
+  allowedlist: ['navigation'] // only navigation will be persisted
 };
 ```
 
 ## Nested Persists
-Nested persist can be useful for including different storage adapters, code splitting, or deep filtering. For example while blacklist and whitelist only work one level deep, but we can use a nested persist to blacklist a deeper value:
+Nested persist can be useful for including different storage adapters, code splitting, or deep filtering. For example while avoidlist and allowedlist only work one level deep, but we can use a nested persist to avoidlist a deeper value:
 ```js
 import { combineReducers } from 'redux'
 import { persistReducer } from 'redux-persist'
@@ -166,13 +166,13 @@ import { authReducer, otherReducer } from './reducers'
 const rootPersistConfig = {
   key: 'root',
   storage: storage,
-  blacklist: ['auth']
+  avoidlist: ['auth']
 }
 
 const authPersistConfig = {
   key: 'auth',
   storage: storage,
-  blacklist: ['somethingTemporary']
+  avoidlist: ['somethingTemporary']
 }
 
 const rootReducer = combineReducers({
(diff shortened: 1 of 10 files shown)


pro-nasa:
Blacklist and whitelist is the best universal term to describe: blocklist/rejectedlist/denylist and allowlist/acceptlist/etc. It's really illogical to associate color names in terms with oppressive language. After all, there are also black scripts/characters/holes...

papb:
> Blacklist and whitelist are the best universal terms to describe: blocklist/rejectedlist/denylist and allowlist/acceptlist/etc.

No.

> It's really illogical to associate color names in terms with oppressive language.

It depends.

> After all, there are also black scripts/characters/holes...

[Faulty generalization](https://en.wikipedia.org/wiki/Faulty_generalization).

Also [here is an interesting short paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/pdf/jmla-106-527.pdf).

papb:
By the way, I didn't mean to be harsh. This is a sensitive topic and rarely the kind of discussion that takes place on GitHub... I think the key point here is to see this change of terminology as a gesture of respect. I also agree that lots of people never really noticed these words as oppressing at all. And of course changing two words in a JavaScript library is immensely far from solving the real problem. However, if this change would make one person that suffers from racism feel slightly better, and make other people think about the topic, then it's already worth it. The tiny effort of some developers that decide to upgrade from `v6` to `v7` (which would be born with this PR) and have to make a few text replacements is, well, tiny (assuming of course that a good migration guide is provided, as it should happen with any breaking change in any library for any reason).

pro-nasa:
@papb The publication you recommended is one big gibberish to me

> > Blacklist and whitelist are the best universal terms to describe: blocklist/rejectedlist/denylist and allowlist/acceptlist/etc.
> 
> I disagree...
> 

You have the right to it. Both opinions are subjective

> > It's really illogical to associate color names in terms with oppressive language.
> 
> It depends.
> 

Yeah, you're right, some other words may be like that, but you unnecessarily paying attention to words, not to the context of the sentence

> > After all, there are also black scripts/characters/holes...
> 
> [Faulty generalization](https://en.wikipedia.org/wiki/Faulty_generalization).
> 

Except black holes (with that term you are right), the rest of the terms for me belong to the same group. The explanation that "black === bad" can also be applied to "black sheep/script/character".

> Also [here is an interesting short paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/pdf/jmla-106-527.pdf).

Thanks for this article, I will read it later

>> However, if this change would make one person that suffers from racism feel slightly better, and make other people think about the topic, then it's already worth it
>

In some specific cases - of course, but this way of thinking is also like a double-edged sword. Note that what I disagree with is assigning these words a new emotional negative / positive meaning associated with a specific race of people. Many words describing, e.g. black people have a negative meaning today. I believe we should try to bring the words back to a neutral tone. For me, this is the right way to create a truly inclusive culture free of racism, xenophobia, stereotypes, etc. Otherwise, freedom of speech will be further restricted.

gabmmontenegro:
Thank you, @sanntanna.

Hi @rt2zz or any contributor, is there a near opportunity for this PR be merged to base branch?

There is also another opened PR with the same purpose: https://github.com/rt2zz/redux-persist/pull/1213

This is a important step to remove offensive terms and promote inclusive language to open source projects.

Thanks in advance.
Gab

pro-nasa:

tincho:
> Also [here is an interesting short paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6148600/pdf/jmla-106-527.pdf).

That paper is no reference at all. It says that when authors use blacklist using quotes, it's because they know it's inappropriate. They use quotes because there is not literal "list" literally painted with the color black. So that's no.1 fallacy here.

It doesn't cite any argument to say that in the black-white dualism, the former means "bad" and the latter "good". Also says that metaphor is traced back to Descartes, or even ancient greeks, that it is "western-based and eurocentric", which is **plain false**. The metaphor has been used also millenia ago in the east. See the famous "yin-yang" symbol, or the games of draughts or chess, where black and white squares represent that duality, light/darkness, day/night, cosmos/chaos, etc.

What it is indeed "western thinking" is to think that one is bad and the other is good. There's no such things. The opposites need each other, they complement. There's yin in yang and there's yang in yin.

And this has absolutely nothing to do with race or skin color.

papb:
@tincho Thanks, you have good points. I will believe your claims rather than spending time verifying them. However, even with all that, I believe you'd agree that the words allowlist and denylist are certainly not *worse* than whitelist and blacklist. Be it oppression or not, racism or not, I'd be in favor of simply refactoring an API to replace a metaphor by precise literal wording. But on top of that there is another benefit, which is to explicitly show respect for Black people. Even if you're right in this particular case and these terms aren't racist at all (and I'm not sure you are, but still), performing this change is still an act of respect for them.

wmonecke:
Will we be implementing also pronouns to our variables `const number: he/him = 5`.
Why try to make something about race when it is just colours? Jesus.

alfonsosn:
> @tincho Thanks, you have good points. I will believe your claims rather than spending time verifying them. However, even with all that, I believe you'd agree that the words allowlist and denylist are certainly not _worse_ than whitelist and blacklist. Be it oppression or not, racism or not, I'd be in favor of simply refactoring an API to replace a metaphor by precise literal wording. But on top of that there is another benefit, which is to explicitly show respect for Black people. Even if you're right in this particular case and these terms aren't racist at all (and I'm not sure you are, but still), performing this change is still an act of respect for them.

Well said. It's actually confusing to use "whitelist" and "blacklist." Allow and Deny is far more clear within a taxonomy perspective.

PR #986, opened 2019-02-14
Add an option to delay the automatic persistence.

leoek:
There are use cases, where the persistence cannot be started immediately. For example, if there is a passphrase from the user needed, to complete the initialization of the storage.

This PR adds the option `manualPersist`, which if set to true, avoids that the persistence is started automatically. `persistor.persist()` can be called later, to start persistence. For example if the initialization of the storage is completed.

The option is optional, therefore there wont be any compatibility issues, with previous versions. If the options object or the property is not set, everything will behave as before.

This fixes #952

PR #1254, opened 2020-10-10
Update types.d.ts

etienne-martin:
Added the `manualPersist` option introduced by https://github.com/rt2zz/redux-persist/pull/986


3imed-jaberi (approved):
lgtm

PR #1261, opened 2020-11-07
fixed bug given config does not have a deserialize prop

EdmundsEcho:
Thank you for this useful utility.  I've been using it for a couple of years now.

I have a storage package with its own serialization/deserialization capacity.  When I turned the `config.serialize: false`, the application failed.  This proposed updated fixes two things:

1. infers the `deserialized` value from the `config.serialize` setting (consistent with both the docs and type annotation)

2. it removes the code that presumes that config.serialize and `config.deserialize` can be a function; inconsistent with the type annotations.

\- E


3imed-jaberi (approved):
sgtm

PR #1282, opened 2021-03-04
improved type for white/black lists of persistConfig

SerhiyZheliznjak:
Forces each value in whitelist and blacklist to be a valid key of root state


3imed-jaberi (approved):
sgtm

PR #1289, opened 2021-04-01
Moving away from the words ‘blacklist’ and ‘whitelist’

adam-sajko:
This topic is :fire:HOT:fire: lately and many large companies like Google moving away from these words to be more inclusive.

This PR marks the 'blacklist' and 'whitelist' properties as deprecated until the next major update. That shouldn't broke anything and should satisfy both sides.

Screenshot from 2021-04-01 15-02-34
Screenshot from 2021-04-01 15-24-08
Screenshot from 2021-04-01 15-24-31

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ const App = () => {
   - arguments
     - [**config**](https://github.com/rt2zz/redux-persist/blob/master/src/types.js#L13-L27) *object*
       - required config: `key, storage`
-      - notable other config: `whitelist, blacklist, version, stateReconciler, debug`
+      - notable other config: `allowlist, blocklist, version, stateReconciler, debug`
     - **reducer** *function*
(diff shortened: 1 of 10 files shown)


pro-nasa:
This PR is just an alternative to the already existing PR, but without increasing the major version. Nothing else.

papb:
I'd recommend `denylist` instead of `blocklist`, I've seen `denylist` used in many places around but haven't seen `blocklist` yet. Just to make it closer to the trend.

1ak31sha (approved):

spacechurro:
@ckalika Any update on updating terminology? I think there are a few open PRs regarding this.

wmonecke:
I think this is not a priority. There are some other critical issues that need to be addressed first.

BeKnowDo:
Do social issues really need to nag us in ever aspect of our lives? It's basically a full time job to not offend anyone. No thanks. That's no way to live.

PR #1324, opened 2021-07-24
If throttle is 0, do not use interval.

cristianoccazinsp:
Fixes persistence in background.

This allows for persistence to fire in RN while in background if throttle is not 0, by removing intervals.


tuandinh0801:
Please merge this

Prosquid1:
This fixes the issue where AsyncStorage does not set data on Android (onPause/onStop/onDestroy) but saves data on iOS.

PR #1332, opened 2021-08-11
feature: resync state

SebastienGllmt:
In our application, we want to resync our redux state from localStorage when the user re-focuses the tab

We do this so that if the user has multiple pages accessing the same localStorage instance, the user tab the user has focused always pulls the latest state. 

In a sense, this `resync` action is the opposite of `flush`

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ const App = () => {
       - pauses persistence
     - `.persist()`
       - resumes persistence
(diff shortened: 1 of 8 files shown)


SebastienGllmt:
Closing in favor of #1403 because the whole repo got rewritten in typescript since I originally made this PR

PR #1369, opened 2022-03-17
Doc: Add explanation of how migrations work

VladimirRubin:
I just added some clarity on how migrations work under the hood to better understand the tool

### DIFF
--- a/docs/migrations.md
+++ b/docs/migrations.md
@@ -1,4 +1,9 @@
-# Redux Persist Migration Example
+
+Redux Persist takes all functions from `migrations` above the inbound persisted store version and runs them in ascending order (by version key) to return the final state
+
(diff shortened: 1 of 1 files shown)


olso:
hey, we've found out the hard way, that https://github.com/rt2zz/redux-persist#state-reconciler also plays a big role

VladimirRubin:
> hey, we've found out the hard way, that https://github.com/rt2zz/redux-persist#state-reconciler also plays a big role

Could you please give more details?

someengineer-nordsec:
> > hey, we've found out the hard way, that https://github.com/rt2zz/redux-persist#state-reconciler also plays a big role
> 
> Could you please give more details?

https://github.com/rt2zz/redux-persist/blob/master/docs/api.md#type-statereconciler

PR #1377, opened 2022-04-19
use `configureStore` in README

phryneas:
Since `redux` 4.2.0, `createStore` is deprecated in favor of `configureStore`. This updates the README to use the recommended function.


phryneas:
Could you please get this merged? It's just a (really necessary) docs update.

yestay90:
@phryneas , thanks for your PR. I had problem with persistStore which threw error `max calls stack exceeded` . The problem was I set persistStore inside of function and  exported function. Later, I used that function inside of useEffect when app loads. After looking at your PR I realized that instead exporting function, I need to export variable. And this solved my error! I feel happy now)) Thanks a lot.

PR #1381, opened 2022-04-29
Fix link to `config` type definitions in `ReadMe`

bent0b0x:
A link in the readme is pointing to the wrong file/line numbers


NikaBuligini:
Looks like the `types` file was updated once again. A new URL can be: `https://github.com/rt2zz/redux-persist/blob/master/types/types.d.ts#L26-L48`

PR #1415, opened 2022-10-25
sap-term-changes: Rename black/white according to SAP

sbimochan:
SAP is dropping the use of master/slave and blacklist/[whitelist](https://www.techtarget.com/whatis/definition/whitelist). It wants its language to be as inclusive as possible and is replacing the well-known terms with alternatives, depending on usage.
Shall we rename blacklist/whitelist terminologies as well?

This would be a breaking change and would require a doc to list down the migration step. I'm happy to edit that page but couldn't find it. It could be written on change log file as well.

more read [here](https://www.techtarget.com/searchsap/news/252488768/SAP-sets-alternatives-to-master-slave-blacklist-whitelist#:~:text=The%20blacklist%2Fwhitelist%20alternatives%20have,its%20push%20for%20term%20changes.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ const App = () => {
   - arguments
     - [**config**](https://github.com/rt2zz/redux-persist/blob/master/src/types.js#L13-L27) *object*
       - required config: `key, storage`
-      - notable other config: `whitelist, blacklist, version, stateReconciler, debug`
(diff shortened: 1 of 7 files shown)


kritish-dhaubanjar:
Agreed! I support this change!

PR #1465, opened 2024-02-02
Replace recommended storage engine for secure storage on react-native

heg2:
The previously recommended `redux-persist-sensitive-storage` only encrypts the data on iOS; on Android it is stored in clear text and readable via the file browser. There exists an alpha-Branch which solves the problem, but this is in alpha years now and not updated anymore. Also, this branch fails on Android 9 and older.

We switched to [react-native-encrypted-storage](https://www.npmjs.com/package/react-native-encrypted-storage) in all our projects; it solves the problems mentioned above and works fine. 

This would be a better suggestion than the currently suggested storage library, which has a high risk for security breaches for Android users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@ const persistConfig = {
   **[redux-persist-indexeddb-storage](https://github.com/machester4/redux-persist-indexeddb-storage)** recommended for web via [localForage](https://github.com/localForage/localForage)
 - **[redux-persist-node-storage](https://github.com/pellejacobs/redux-persist-node-storage)** for use in nodejs environments.
 - **[redux-persist-pouchdb](https://github.com/yanick/redux-persist-pouchdb)** Storage engine for PouchDB.
-- **[redux-persist-sensitive-storage](https://github.com/CodingZeal/redux-persist-sensitive-storage)** react-native, for sensitive information (uses [react-native-sensitive-info](https://github.com/mCodex/react-native-sensitive-info)).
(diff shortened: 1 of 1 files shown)


losh11:
FYI react-native-encrypted-storage has been deprecated. It hasn't been maintained in over a year, and has been archived by the repo creators.

heg2:
I can't find any notes of it being deprecated or archived? 
While it's true that it has not been updated for a year, it is a working solution (unlike the currently recommended redux-persist-sensitive-storage, which by the way has not been updated for seven years).

However, this repo here seems pretty abandoned too, so... 🤷

losh11:
If you open up the encrypted-storage repo on [github](https://github.com/emeraldsanto/react-native-encrypted-storage), you can see there's a banner on the top that says that the repo has been archived.

evanwalsh:
FWIW, this is the thin wrapper around `react-native-keychain` that I use:

```typescript
import {
  getGenericPassword,
  resetGenericPassword,
  setGenericPassword,
} from 'react-native-keychain'

const KeychainPersistStore = {
  async getItem(key: string): Promise<string | undefined> {
    const value = await getGenericPassword({service: key})

    if (value) {
      return value.password
    }
  },
  async setItem(key: string, value: string): Promise<void> {
    await setGenericPassword('data', value, {service: key})
  },
  async removeItem(key: string): Promise<void> {
    await resetGenericPassword({service: key})
  },
}
```

